PR #7, opened 2019-02-08
Added some basic Lua examples & fixed some dead links

Jarrrk:


content/scripting-reference/runtimes/lua/functions/GetPlayerIdentifiers.md
```diff
@@ -8,10 +8,22 @@ Syntax
...
+```
+steam:110000132ca2xxx, license:cac6120ec931e564afae342956b9742d4d9aexxx, xbl:2535439919044xxx, live:985157396007xxx, ip:95.45.101.xxx
+```
+These can be used to identify a player based on either their steam account, hardware identifier, XBL/GFWL identifier and also IP address.
+
+For example, to just get the players steam identifier you can do the following:
+```lua
+local playerID = 1 -- The player ID to return identifiers for
+
+local steamID = GetPlayerIdentifiers(playerID)[1]
```
stannum-cfx: This is wrong. You can't rely on the steam identifier being at index 1, it's not guaranteed to be present or at a specific index. You actually have to loop through the identifiers and find the one containing `steam:`.

content/scripting-reference/runtimes/lua/functions/GetPlayerIdentifiers.md
```diff
@@ -8,10 +8,22 @@ Syntax
...
 ```lua
-GetPlayerIdentifiers()
+GetPlayerIdentifiers(int playerID)
 ```
 
 Examples
 --------
+An example return table of the GetPlayerIdentifiers currently looks like the following:
+```
+steam:110000132ca2xxx, license:cac6120ec931e564afae342956b9742d4d9aexxx, xbl:2535439919044xxx, live:985157396007xxx, ip:95.45.101.xxx
```
stannum-cfx: What about discord?

cerium-cfx:
Changes were requested, but not resolved over the past 3 months. Considering this PR as stale.

PR #11, opened 2019-03-06
blips: add blip colors to blips page

PichotM:


blattersturm:
Are these different from the normal HUD colors?

PichotM:
I believe yes, since it differs a lot
https://pastebin.com/d9aHPbXN HUD colors
https://gtaforums.com/topic/864881-all-blip-color-ids-pictured/ blip colors

blattersturm:
Ah - blip colors are mapped using some sort of mapping table to HUD colors, which is defined in the EXE.

In addition to that, certain blip colors actually aren't _constant_ but depend on another flag - probably friendly/enemy flag:

- 2, 69
- 3, 67, 68
- 4
- 5, 66
- 49

Would it be possible to gather/add both variations of these ones?

PichotM:
yeahh, by using SET_BLIP_AS_FRIENDLY my findings:
enemy - default - friendly
2: red green blue
3: red blue blue 
4: red white blue 
5: red yellow blue 
49: red red blue 
66: red yellow blue 
67: red blue blue 
68: red blue blue 
69: red green blue 

Any idea on how this could be implemented in the blips page?

stannum-cfx:
@PichotM create a new PR for the different variations once @blattersturm replies. I've merged it in the meantime because it's still useful for now.

PR #13, opened 2019-03-16
playerConnecting: add a lua example

PichotM:
Warxander asked :(

### DIFF
--- a/content/scripting-reference/events/list/playerConnecting.md
+++ b/content/scripting-reference/events/list/playerConnecting.md
@@ -55,14 +55,19 @@ private void OnPlayerConnecting([FromSource]Player player, string playerName, dy
 ##### Lua Example:
 ```lua
 local function OnPlayerConnecting(name, setKickReason, deferrals)
-    local identifiers = GetPlayerIdentifiers(source)
+    local identifiers, steamIdentifier = GetPlayerIdentifiers(source)
     deferrals.defer()
 
     deferrals.update(string.format("Hello %s. Your steam id is being checked.", name))
-    Citizen.Wait(1000) -- not needed, we fake a loading
 
-    local steam = identifiers[1] -- The steam identifier is always the first identifier (if steam is enabled)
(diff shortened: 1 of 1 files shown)


warxander:
Actually, my question was about `deferrals` as I didn't use them before and, imo, the current description is not clean enough.
And I also confused about `setKickReason` because it's not used in C# example.

PichotM:
deferrals has been implemented after setKickReason function
according to iridium, it is better to use deferrals than setKickReason callback 👍

content/scripting-reference/events/list/playerConnecting.md
```diff
@@ -51,3 +51,23 @@ private void OnPlayerConnecting([FromSource]Player player, string playerName, dy
...
 ```
+
+##### Lua Example:
+```lua
+local function OnPlayerConnecting(name, setKickReason, deferrals)
+    local identifiers = GetPlayerIdentifiers(source)
+    deferrals.defer()
+
+    deferrals.update(string.format("Hello %s. Your steam id is being checked.", name))
+    Citizen.Wait(1000) -- not needed, we fake a loading
```
stannum-cfx: Why even add this delay if it's not needed. People tend to just blindly copy code snippets like these and then everyone will have a useless delay when entering the server, doesn't seem very user friendly to me.

content/scripting-reference/events/list/playerConnecting.md
```diff
@@ -51,3 +51,23 @@ private void OnPlayerConnecting([FromSource]Player player, string playerName, dy
...
+
+    local steam = identifiers[1] -- The steam identifier is always the first identifier (if steam is enabled)
```
stannum-cfx: You can't just assume that it's at any specific index, if there's any identifier related changes in the future, this _could_ potentially change. It's not recommended to do it like this. Please just loop over all of them to check if _any_ identifier is the steam ID.

PichotM:
It should be better now

PR #15, opened 2019-03-21
blips: add friendly-enemy color variations

PichotM:
only colors that differ from the default one are added


stannum-cfx:
Looks good, I'd just add a small description below the Blip colors header explaining what the `(enemy)` and `(friendly)` variations are and how you can use them.

PichotM:
Obviously, what I forgot

PichotM:
Added!

PR #16, opened 2019-03-21
Improve setup instructions

cerium-cfx:
Specify theme supported version of hugo - 0.49 - as theme relies on *Source* struct that was removed in 0.50

See https://github.com/vjeantet/hugo-theme-docdock/issues/166


stannum-cfx:
Docs theme has been updated to solve this, you'll no longer need a specific version.

PR #24, opened 2019-04-17
Add sv_auth convars/commands to server commands page

jakenotjacob:
Update the server commands section to include an explanation of the sv_auth{Variance,Trust} convars/commands.


AwakenDev:
That's smart to add

jakenotjacob:
I held off adding a shortened description directly above the convars in the example server.cfg to follow prior comments, though it _seems_ like it may be a good idea to avoid folks having to dig further into the docs to find a summary? Thoughts?

PR #30, opened 2019-05-02
Update listening-for-events.md

ioerrorgit:
Added section about 'source' and small security note.

### DIFF
--- a/content/scripting-manual/working-with-events/listening-for-events.md
+++ b/content/scripting-manual/working-with-events/listening-for-events.md
@@ -77,3 +77,24 @@ onNet('eventName', (eventParam1, eventParam2) => {
 ```
 
 _You only need to register events when using Lua or JS -- C# does not require you to do this manually, though you might want to protect your server events with own code._
+
+#### Determining the source of a server event
+When we register an event on the server side, and make it available over network with RegisterServerEvent, it is useful to determine the client that **called** that event.  We can do this by using the built in SOURCE argument, that is passed whenever a Server Event is triggered.  Source is the client which triggered the event, and while completely invisible at first glance, it is always passed.
+
+**Example**
+```lua
+Client.lua
+TriggerServerEvent("eventName")
+
+Server.lua
+
+RegisterServerEvent("eventName")
+AddEventHandler("eventName", function()
+print(source) -- will print the ID of the client which called the event
+end)
+```
+
+
+#### Note about event security
+
+Keep in mind that while events can be extremely powerful and useful over the network, some general security practices should be acknowledged.  When you register a networked server event, you are effectively allowing **anyone** to call it.  By anyone, I mean any server or client resource - however, should an exploiter have access to a lua executor they can trigger these too.  Be cautious as to how you allow server events to be executed.  For instance - implementing permissions checks on the server side within the event, instead of checking only on the client script would be a good security practice.


content/scripting-manual/working-with-events/listening-for-events.md
```diff
@@ -77,3 +77,24 @@ onNet('eventName', (eventParam1, eventParam2) => {
 ```
 
 _You only need to register events when using Lua or JS -- C# does not require you to do this manually, though you might want to protect your server events with own code._
+
+#### Determining the source of a server event
+When we register an event on the server side, and make it available over network with RegisterServerEvent, it is useful to determine the client that **called** that event.  We can do this by using the built in SOURCE argument, that is passed whenever a Server Event is triggered.  Source is the client which triggered the event, and while completely invisible at first glance, it is always passed.
```
cerium-cfx: Unnecessary capitalisation of `Server Event`.

content/scripting-manual/working-with-events/listening-for-events.md
```diff
@@ -77,3 +77,24 @@ onNet('eventName', (eventParam1, eventParam2) => {
...
+
+**Example**
+```lua
+Client.lua
```
cerium-cfx: This should be a code comment

content/scripting-manual/working-with-events/listening-for-events.md
cerium-cfx: Depending on the scripting language, it is not always `RegisterServerEvent`. This is very specific to Lua scripting. Perhaps omit `with RegisterServerEvent` or explain that this is language dependent.

content/scripting-manual/working-with-events/listening-for-events.md
```diff
@@ -77,3 +77,24 @@ onNet('eventName', (eventParam1, eventParam2) => {
...
+TriggerServerEvent("eventName")
+
+Server.lua
```
cerium-cfx: Same as above

content/scripting-manual/working-with-events/listening-for-events.md
```diff
@@ -77,3 +77,24 @@ onNet('eventName', (eventParam1, eventParam2) => {
...
+
+RegisterServerEvent("eventName")
+AddEventHandler("eventName", function()
+print(source) -- will print the ID of the client which called the event
```
cerium-cfx: Indent?

content/scripting-manual/working-with-events/listening-for-events.md
```diff
@@ -77,3 +77,24 @@ onNet('eventName', (eventParam1, eventParam2) => {
...
+end)
+```
+
+
+#### Note about event security
+
+Keep in mind that while events can be extremely powerful and useful over the network, some general security practices should be acknowledged.  When you register a networked server event, you are effectively allowing **anyone** to call it.  By anyone, I mean any server or client resource - however, should an exploiter have access to a lua executor they can trigger these too.  Be cautious as to how you allow server events to be executed.  For instance - implementing permissions checks on the server side within the event, instead of checking only on the client script would be a good security practice.
```
cerium-cfx: Please don't explicitly mention exploit methods (if available at the time). The user just has to be informed of the security risks associated with networked events.

cerium-cfx:
Stale PR

PR #33, opened 2019-05-02
Update disabling-antivirus.md

strontuim:


cerium-cfx:
Bad grammar. Adding links for adding exceptions to other anti-virus software is preferred over sending users to Google.

PR #35, opened 2019-05-11
Update disabling-antivirus.md

strontuim:
added windows defender because it also cause some problem

### DIFF
--- a/content/client-manual/disabling-antivirus.md
+++ b/content/client-manual/disabling-antivirus.md
@@ -17,4 +17,4 @@ solution for these issues.
 - Add it to your antivirus exclusions. For the common antivirus vendors:
     - [Avast](https://support.avast.com/en-eu/article/Antivirus-scan-exclusions)
     - [AVG](https://support.avg.com/SupportArticleView?urlname=How-to-exclude-file-folder-or-website-from-AVG-scanning)
-    - [windows defender](https://www.wikihow.com/Turn-Off-Windows-Defender-in-Windows-10)
+    - [Windows Defender](https://www.wikihow.com/Turn-Off-Windows-Defender-in-Windows-10)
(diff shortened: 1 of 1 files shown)


content/client-manual/disabling-antivirus.md
```diff
@@ -17,3 +17,4 @@ solution for these issues.
 - Add it to your antivirus exclusions. For the common antivirus vendors:
     - [Avast](https://support.avast.com/en-eu/article/Antivirus-scan-exclusions)
     - [AVG](https://support.avg.com/SupportArticleView?urlname=How-to-exclude-file-folder-or-website-from-AVG-scanning)
+    - [windows defender](https://www.wikihow.com/Turn-Off-Windows-Defender-in-Windows-10)
```
JHodgson1: `Windows Defender` would be grammatically correct.

strontuim: thanks :p

BlackFlash5: Don't turn it off, just add an exclusion.
Also link to official documentation from Mircosoft, not some 3rd party.

e.g.: https://support.microsoft.com/en-us/help/4028485/windows-10-add-an-exclusion-to-windows-security

strontuim: exclusion cant help in some cases. thats why disabling is better.

JHodgson1: This is typically only the case when dealing with third-party antivirus. Same case with disabling -- commonly, third-party antiviruses don't actually disable when the user specifically says that it wants it disabled. So to be sure, uninstalling is probably a better solution for third-party antiviruses.

BlackFlash5: Before i would disable Windows Defender, i would upload the file to Microsoft for a review (takes normally about 24-48h), so they can check the file and update the signature.
This way you don't have to disable your AV and everyone else gets also the updated signature.
This is probably not a practical solution if thats the case after every update, but even then you should show both options (exclusion and disabling) and not only the disabling one.

https://www.microsoft.com/en-us/wdsi/filesubmission

strontuim: well fivem is safe and we are not talking a bout just a file. we are talking about a folder. all you are saying make no sense. thanks. if you dont trust fivem dont use it. million of people are using it currently and its fine so......

strontuim: deleted it by mistake enjoy :(

PR #46, opened 2019-06-23
Gamer Tags changes

balika011:


balika011:
depends on https://github.com/citizenfx/natives/pull/133

PR #57, opened 2019-07-11
game-references: ped models

PichotM:
⚠️

This PR is quite heavy.
If you want "thumbnails" or a lower quality just let me know

Also, I don't know how or if it is possible to "not resize" an image in the 'onclick' featherlight.


PichotM:
Preview https://i.imgur.com/s70PSpu.png

cerium-cfx:
Looks good. I don't think the size of the images will be an issue. If it is, is it even possible to use "thumbnails"?

PichotM:
yes I can generate thumbnails the "featherlight" feature will be then unavailable and clicking on the thumbnail will lead to the full image

PR #62, opened 2019-07-27
server-installation: Fix the ambiguity in some of the steps

Zeemahh:
I felt as if some of these steps needed further clarification to make sense with the steps mentioned later on in the guide, some wording improvements were added and changing the directory for the example


cerium-cfx:
What about the Linux installation steps?

PR #65, opened 2019-08-28
Update setting-up-a-server.md

Local9:
Clearing up a couple of points to aid in reading.
Adding the `call` cmd command as this I hope can reduce confusion.

If required I can also add images.


cerium-cfx:
Pull request https://github.com/citizenfx/fivem-docs/pull/62 was accepted instead of this. It was a bit more elaborate and an older PR.

PR #66, opened 2019-09-20
Procdump: -e 1 will die on first error

C0kkie:
If you set the parameter to 1 it will write instantly a log, if you remove the "1" it will write a log if server will crash


cerium-cfx:
Stale PR

PR #69, opened 2019-11-23
Removed reference to deleted Discord channel

HaxersAlwaysWin:


rex2630:
WTF? That discrod does still exists...!

manups4e:
i think this is mainly to mitigate all those same questions that are made everyday (every 10 mins i'd add)
and in this way using the forum they could notice some threads with maybe the same question and already answered.

hojgr:
idk, dont like it. While we are at it, the symbols link is dead - https://runtime.fivem.net/client/symbols/

Something-Debug:
What do you mean the symbols link is dead? It works as intended.

manups4e:
he means this 
image

Disquse:
> he means this
This is fine because that link is supposed to be used in IDEs which generating full path to symbol files.

cerium-cfx:
The Discord invite link still works. We can update it with the vanity URL now though. As for the channel, yes that doesn't exist anymore but instead if called `#fxserver-support`.

If you'd like, update the PR with the info given above please.

PR #84, opened 2020-02-13
Update submodule installation instructions

Jamelele:
Following the current instructions left me a bit confused until I figured out that the nested submodules weren't cloned.


blattersturm:
Nested submodules are not required however.

blattersturm:
Right, this requirement might've changed when switching to `docsy`.

PR #123, opened 2020-05-16
Spelling

NinjaLabs-Dev:
Fixed spelling of dependencies


Jamelele:
It's not a typo

https://github.com/citizenfx/fivem/blob/fd3fae946163e8af472b7f739aed6f29eae8105f/code/components/citizen-resources-core/src/ResourceDependencyLoader.cpp#L58

Just an alias of 'dependency' which allows you to use the correctly spelled plural, otherwise your only option for pluralisation would be

```lua
dependencys {
   ...
}
```

PR #125, opened 2020-05-22
AddEventHandler examples and fixed invalid url

Poggicek:


content/docs/scripting-reference/runtimes/lua/functions/AddEventHandler.md
```diff
@@ -18,6 +18,18 @@ AddEventHandler(string eventName, function callback)
 Examples
 --------
 
-TODO
+-- SERVER
 
-[listening-for-events]: /guides/listening-for-events
+Don't forget to [RegisterNetEvent][]!
```
Zeemahh, line 23: Should add a reference to https://github.com/citizenfx/fivem-docs/blob/master/content/docs/scripting-manual/working-with-events/listening-for-events.md#note-when-using-luajs here (of course change with an actual path for proper redirecting), instead of no URL.

Issue #120, opened 2020-05-11
Error building site: failed to render shortcode "native_link"


Jamelele:
Getting this error when building the docs, even on the master branch with no edits, I'm not quite sure why.

```
$ hugo server -D
Building sites … Built in 2780 ms
Error: Error building site: "D:\Jamelele\Libraries\Documents\Code\fivem-docs\content\docs\developers\script-runtimes.md:170:54": failed to render shortcode "native_link": failed to process shortcode: "D:\Jamelele\Libraries\Documents\Code\fivem-docs\layouts\shortcodes\native_link.html:14:72": execute of template failed: template: shortcodes/native_link.html:14:72: executing "shortcodes/native_link.html" at <$native.hash>: can't evaluate field hash in type int
```

```
$ hugo version
Hugo Static Site Generator v0.67.1/extended windows/amd64 BuildDate: unknown
```

PR #129, opened 2020-05-23
fix: "can't evaluate field hash in type int"

Zeemahh:
Fixes building failing when specifying `hash` index directly.

~~Though, because of the removal of the `CFX` namespace at `https://runtime.fivem.net/doc/natives.json`, any `CFX` native passed to `<native_link>` or `<native>`, it will just return `https://runtime.fivem.net/doc/natives/?_0` - as `0` is the default value for `$nativeHash`. Without this, the build will fail.~~

Resolves https://github.com/citizenfx/fivem-docs/issues/120


Zeemahh:
> any CFX native passed to <native_link> or <native>, it will just return https://runtime.fivem.net/doc/natives/?_0

Fixed in [`5371f25`](https://github.com/citizenfx/fivem-docs/pull/129/commits/5371f255ef1cdcfecacbc1c08513a3f6c549d1d4).

PR #134, opened 2020-05-26
GetPlayerIdentifiers changes.

Poggicek:


content/docs/scripting-reference/runtimes/lua/functions/GetPlayerIdentifiers.md
```diff
@@ -4,6 +4,22 @@ title: GetPlayerIdentifiers
 
 Returns a table containing all of the player’s identifiers.
 
+**Never try to get identifiers based on their index. Identifier positions MAY vary!!!!**
```
Zeemahh, line 7: >MAY

Should be 'will'.

content/docs/scripting-reference/runtimes/lua/functions/GetPlayerIdentifiers.md
```diff
@@ -17,32 +33,23 @@ GetPlayerIdentifiers(Player player)
...
-  end
+function GetAllIdentifiers(player)
+    local identifiers = {}
+    for i = 0, GetNumPlayerIdentifiers(player) - 1 do
+        local raw = GetPlayerIdentifier(player, i)
+        local source, value = raw:match("^([^:]+):(.+)$")
+        if source and value then
+            identifiers[source] = value
+        end
+    end
```
Zeemahh, line 49:
```suggestion
for key, raw in pairs(GetPlayerIdentifiers(player)) do
    local source, value = raw:match("^([^:]+):(.+)$")
    if source and value then
        identifiers[source] = value
    end
end
```

CryptoGenics: pairs/ipairs are about 2x as slow as i=

Good the way it is 😀

fredrik-krokedil: Yes, it's fine the way it is, but the `for i` variant doesn't actually include `GetPlayerIdentifiers`, and that's what this is supposed to document the usage of.

cerium-cfx:
Creating the `GetAllIdentifiers` function is not necessary. In the Lua scheduler `GetPlayerIdentifiers` exists. See [here](https://github.com/citizenfx/fivem/blob/d911ecf638337c7c61fc6728110c92d84a217156/data/shared/citizen/scripting/lua/scheduler.lua#L427-L436)

PR #135, opened 2020-05-26
TriggerEvent example

Poggicek:


content/docs/scripting-reference/runtimes/lua/functions/TriggerEvent.md
```diff
@@ -20,4 +20,14 @@ TriggerEvent(string eventName[, ...])
...
 --------
 
-TODO
+Client
+```lua
+TriggerEvent('eventName', 'Hello world!')
+```
+Server
+```lua
+RegisterNetEvent('eventName')
```
Zeemahh, line 29: You don't need to register as net event if you're not using it over the net, i.e. client->server/vice versa.

Poggicek: Oh thats actually new for me. It does make sense

Poggicek: fixed

cerium-cfx:
`TriggerEvent` is for events within the same domain (e.g. client), so this example does not work. Both `AddEventHandler` and `TriggerEvent` must be on either the client or server. Please fix this

technetium-cfx:
Please, test your examples. Triggering an event before it exists has no effect.

PR #136, opened 2020-05-27
Add weaponDamageEvent lua example

arthur-rl:


cerium-cfx:
Personally I find the examples confusing and the explanation at the end unclear.

arthur-rl:
Hopefully updated to make the example more clear.

PR #154, opened 2020-07-01
Update server.cfg

realtkco:
Not sure if that was a accident or what?


smallo92:
baseevents is not needed to run a server and is only an example resource

realtkco:
oh.... oops

PR #167, opened 2020-07-24
Update setting-up-a-server.md

abhiss:
Remove implied dependency on WinRAR or 7-Zip in docs. Expand-Archive is a default cmdlet in PowerShell 5.1.


blattersturm:
At the time this was written, the server archive wasn't a `.zip`.

Also, Windows ships with a `tar.exe` that can extract .zip (and I believe some types of `.rar`) since ages.

Additionally, people have independent thought and it's 'such as'?

cerium-cfx:
The tools mentioned are just examples. There's no need for a third example.

PR #171, opened 2020-08-05
Fixing broken links

Hernior:
Modified links to documentation that already exists


cerium-cfx:
Can you please provide relative links and not hardcoded as the docs website url may change.

PR #90, opened 2020-03-09
Added FAQ reference to #NoBans, which is way down at the bottom of th…

jonathonbarton:
#88 Added reference to #nobans from the homepage to the FAQ docs.
If you already know about FiveM or hear about it from a friend, #nobans is an easy thing to miss because it's below the comparisons way at the bottom of the page.
Adding it to the FAQ is a layer of idiot proofing.
Pull request submitted by one of said idiots. :D

Issue #88, opened 2020-03-08
Client FAQ


jonathonbarton:
Q: How do I switch back and forth between playing on FiveM servers and regular Rockstar GTA:O play? Do I have to choose one or the other? Is FiveM counted by R* as a mod? **Will having FiveM installed get me banned in GTA:O**?

A: (I dunno, I've looked and looked and not found the answer - not even in the FAQ...)

Zeemahh:
What? This isn't a place to ask these questions, this is a place for issues regarding the docs. FiveM and GTA:O are totally separate.

jonathonbarton:
It's an issue in the docs because it's a frequently asked question that isn't answered in the FAQ.
It's a missing documentation issue.

nathanctech:
The answer is no, you will not be banned. The game doesn't touch your GTA V install and you can freely play both.

cerium-cfx:
This is already listed on our main website that users visit before downloading the client https://fivem.net/#no-bans

You're more than welcome to PR anything to the docs and we will take a look, but please use the forum or Discord in the future if you cannot find answers to questions.

PR #175, opened 2020-08-13
Added FAQ reference to #NoBans

rjmxtechnology:
which is way down at the bottom of the homepage. If you already know about FiveM, it's easy to miss...

https://github.com/citizenfx/fivem-docs/issues/88 
Continuation of https://github.com/citizenfx/fivem-docs/pull/90 since he closed it for some reason instead of fixing it???

### DIFF
--- a/content/docs/support/client-faq.md
+++ b/content/docs/support/client-faq.md
@@ -26,10 +26,10 @@ That depends on how you installed it in the first place.
 
 More about installing FiveM [here][installing-fivem].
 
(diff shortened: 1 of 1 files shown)


technetium-cfx:
Try to keep the title short and concise. Asking _four_ questions in a single title is a bit much, one should be enough 😉 
Also, I suggest to change `In case you missed it` with `As mentioned on fivem.net`.

rjmxtechnology:
Made suggested changes 👍

PR #178, opened 2020-08-20
removed unnecessary +exec

PLOKMJNB:


arthur-rl:
That's not unnecessary though, it's required, unless you want to launch with txAdmin.

technetium-cfx:
@arthur-rl is correct here, but I am assuming you knew that as well. Didn't mean to reopen, so I'm keeping it closed for now.

This docs article was written far prior to txAdmin. Preferably these instructions are properly rewritten to accommodate usage both with and without txAdmin. That would leave our documentation more complete than just adding or removing a parameter. Feel free to rewrite this section and open a new PR. We welcome easier instructions, including an explanation of launch parameters 😄 

(PS, we do encourage the use of txAdmin, so this PR isn't necessarily wrong 😉)

PR #182, opened 2020-08-23
Document `con_miniconChannels`

sharkykh:
~I'm not sure about the examples, maybe they are *too* detailed.~
Updated.

### DIFF
--- a/content/docs/client-manual/console-commands.md
+++ b/content/docs/client-manual/console-commands.md
@@ -13,6 +13,24 @@ tools if preferred, like [VConsole2][vconsole]. This allows you to use the clien
 The `cmdlist` command will list all the commands that are registered on the client (or server). It will also output the
 variables that have been set by using the `set`, `sets` and `seta` commands.
 
+### con_miniconChannels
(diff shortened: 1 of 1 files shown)


content/docs/client-manual/console-commands.md
```diff
@@ -13,6 +13,28 @@ tools if preferred, like [VConsole2][vconsole]. This allows you to use the clien
 The `cmdlist` command will list all the commands that are registered on the client (or server). It will also output the
 variables that have been set by using the `set`, `sets` and `seta` commands.
 
+### con_miniconChannels
+You can use the `con_miniconChannels` to display console messages on screen without needing to open the client console.
+
+The pattern is matched against channel names. By default, it is set to `minicon:*`.
```
sharkykh, line 19: The pattern is full-matched (equivalent to `^pattern$`) against the channel name.
This means if the pattern is `script:banking`, and an NUI frame from `script:banking` prints something, it will not show unless the pattern is changed to `script:banking*` or `script:banking script:banking:*`.

hoyame: ok

Zeemahh (approved):

PR #192, opened 2020-10-11
Added the explanation of server events and events in general

wowjeeez:


AvarianKnight:
As mentioned in https://github.com/citizenfx/fivem-docs/pull/201 this entire section should be rewritten to complement each other.

PR #201, opened 2020-11-09
Added a server script

stebbzor:
Two simple commands that'll show usage of the GetPlayers() function on the server.


stebbzor (commented):
Server-side script that demonstrates simple usage of the GetPlayers() function on the server, with a little bit of logic. 

- Second revision I've fixed the links I put in to look better.

technetium-cfx (commented):
I personally believe it would make more sense to rewrite this article in such a way that the client and server scripts complement each other. Perhaps something more basic too. There's a lot to grasp here for beginners, and it doesn't quite highlight how server and client scripts are used together.

AvarianKnight:
Closing this based off of the previous comment, this will be rewritten in the future.

PR #202, opened 2020-11-09
Added secondary option for OneSync Infinity

stebbzor:
I believe people will look first to the docs when getting started, and having OneSync Inifinity as an option there will perhaps encourage people to use that. 
Maybe we should consider setting up a detailed page on OneSync on the docs, if there isn't one already.


tabarra:
At this point, this page should already have an txAdmin tutorial, since it's the easiest way to get your fxserver up and running.

PR #212, opened 2020-12-15
Add globbing support alert to file directive

niekschoemaker:
Appeared to be missing in the docs, is supported in the code since https://github.com/citizenfx/fivem/commit/6f271aaf20e89fc4adf99c25d6e6965bf67306f5

### DIFF
--- a/content/docs/scripting-reference/resource-manifest/resource-manifest.md
+++ b/content/docs/scripting-reference/resource-manifest/resource-manifest.md
@@ -219,6 +219,8 @@ file 'html/loadscreen.html'
 
 ### file
 
+{{% alert color="success" title="Note" %}}This directive supports globbing.{{% /alert %}}
+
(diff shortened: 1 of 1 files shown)


niekschoemaker:
Might not want to actually, it appears to be only partly supported.
Had a lot of server crashes whenever I applied this to the file entries.

blattersturm:
.. and why don't you report said crashes with reproducible steps?

technetium-cfx:
@niekschoemaker Would you be able to explain what you used that caused a server crash? Sounds like an issue that needs to be resolved, as crashes are not intended.

niekschoemaker:
Yes I will take a look at that, had some other issues to take care of.

Finally managed to get memory dumps working in docker/linux (was getting blocked by apparmor on linux) so will try to get some dumps from this crash.

technetium-cfx:
An example resource that crashes would be better. We can find the issue more easily if we can reproduce the crash on our end.

PR #219, opened 2020-12-29
Add Lua example for chat:addSuggestions event

encorecharles:
This event could really use an example as the existing documentation doesn't specify that each suggestion object should be an associative table rather than indexed.

### DIFF
--- a/content/docs/resources/chat/events/chat-addSuggestions.md
+++ b/content/docs/resources/chat/events/chat-addSuggestions.md
@@ -23,4 +23,27 @@ array suggestions
 Examples
 --------
 
-TODO
+This example adds a command suggestion for the `/command` and `/othercommand` commands.
+
+##### Lua Example:
+```lua
+-- Note, the command has to start with `/`.
+
+TriggerEvent('chat:addSuggestions', {
+    {
+        name='/command',
+        help='help text',
+        params={
+            { name="paramName1", help="param description 1" },
+            { name="paramName2", help="param description 2" }
+        }
+    },
+    {
+        name='/othercommand',
+        help='other help text',
+        params={
+            { name="otherParamName1", help="other param description 1" },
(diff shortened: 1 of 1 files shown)


content/docs/resources/chat/events/chat-addSuggestions.md
```diff
@@ -23,4 +23,27 @@ array suggestions
...
\ No newline at end of file
+This example adds a command suggestion for the `/command` and `/othercommand` commands.
+
+##### Lua Example:
+```lua
+-- Note, the command has to start with `/`.
+
+TriggerEvent('chat:addSuggestions', {
+    {
+        name='/command',
```
AvarianKnight, line 34: I'm pretty sure you don't need the / to suggest a command, they're done by regular name?

lze3: > I'm pretty sure you don't need the / to suggest a command, they're done by regular name?

No, you do need it.

4mmonium: Yeah, it looks like the slash is actually needed, you can check this out for confirmation 🙂

**Adds the events and sends them to the client:**
https://github.com/citizenfx/cfx-server-data/blob/0210252be7c17745fa30220e12c19e0d2f19af02/resources/%5Bgameplay%5D/chat/sv_chat.lua#L253

**Adds the suggestion on the client:**
https://github.com/citizenfx/cfx-server-data/blob/0210252be7c17745fa30220e12c19e0d2f19af02/resources/%5Bgameplay%5D/chat/cl_chat.lua#L72

Ultimately, this looks OK to me, I haven't tested this example code though, but I believe it should be fine.

4mmonium (approved):
Tested changes by triggering the event on the client, LGTM!

PR #223, opened 2021-01-04
Updated to use GetActivePlayers

Sampleeeee:


lze3 (commented):
Is this code equivalent to the code which is actually shipped as part of cfx-server-data? If so, it's probably more ideal to provide a permalink to the appropriate file in the repository, so we don't have to keep updating this here.

Why did you remove the condition for removing entries which would be unused, also? You changed the condition to be moved inside the code which initialises a gamer tag, which will make that section run always, that's not exactly performant.

PR #227, opened 2021-01-09
Update creating-your-first-script-csharp.md

SaSiNO97:
Changed the complicated `RegisterCommand` with the more readable, newbie-friendly, and overall better `CommandAttribute`.
I tested the new code, it compiles and it works as intended.

### DIFF
--- a/content/docs/scripting-manual/introduction/creating-your-first-script-csharp.md
+++ b/content/docs/scripting-manual/introduction/creating-your-first-script-csharp.md
@@ -33,9 +33,9 @@ namespace MyResourceNameClient
 Easy right? But what about adding functionality? We will start by adding a command using various FiveM scripting concepts.
 
 ```csharp
-using System;
-using System.Collections.Generic;
 using CitizenFX.Core;
+using System;
+using System.Linq;
 using static CitizenFX.Core.Native.API;
 
 namespace MyResourceNameClient
@@ -49,17 +49,20 @@ namespace MyResourceNameClient
 
         private void OnClientResourceStart(string resourceName)
         {
-            if (GetCurrentResourceName() != resourceName) return;
-
-            RegisterCommand("car", new Action<int, List<object>, string>((source, args, raw) =>
+            if (GetCurrentResourceName() != resourceName) 
+                return;
+                
+            // Implement here your resource startup logic
+        }
+        
+        [Command("car")]
+        private void CarCommand(string[] args)
+        {
+            TriggerEvent("chat:addMessage", new
             {
-                // TODO: make a vehicle! fun!
-                TriggerEvent("chat:addMessage", new
-                {
-                    color = new[] {255, 0, 0},
-                    args = new[] {"[CarSpawner]", $"I wish I could spawn this {(args.Count > 0 ? $"{args[0]} or" : "")} adder but my owner was too lazy. :("}
-                });
-            }), false);
+                color = new[] { 255, 0, 0 },
+                args = new[] { "[CarSpawner]", $"I wish I could spawn this {(args.Count() > 0 ? $"{args[0]} or" : "")} adder but my owner was too lazy. :(" }
+            });
         }
     }
 }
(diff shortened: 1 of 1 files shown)


jaymo1011:
Why is the `OnClientResourceStart` event handler still registered manually instead of using `EventHandlerAttribute` or even in the code in the first place when it isn't used by the car spawner logic?

content/docs/scripting-manual/introduction/creating-your-first-script-csharp.md
```diff
@@ -49,17 +49,20 @@ namespace MyResourceNameClient
...
-            if (GetCurrentResourceName() != resourceName) return;
-
-            RegisterCommand("car", new Action<int, List<object>, string>((source, args, raw) =>
+            if (GetCurrentResourceName() != resourceName) 
+                return;
+                
+            // Implement here your resource startup logic
+        }
+        
+        [Command("car")]
```
jaymo1011, line 58: Though not required, it might be a good idea to explicitly add the `Restricted = false` on the attribute to give a hint on how one would restrict usage of the command using the attribute context.

content/docs/scripting-manual/introduction/creating-your-first-script-csharp.md
```diff
@@ -73,36 +76,25 @@ EventHandlers["onClientResourceStart"] += new Action<string>(OnClientResourceSta
...
+[Command("car")]
+private void CarCommand(string[] args)
 {
-    // TODO: make a vehicle! fun!
     TriggerEvent("chat:addMessage", new
     {
-        color = new[] {255, 0, 0},
-        args = new[] {"[CarSpawner]", $"I wish I could spawn this {(args.Count > 0 ? $"{args[0]} or" : "")} adder but my owner was too lazy. :("}
+        color = new[] { 255, 0, 0 },
+        args = new[] { "[CarSpawner]", $"I wish I could spawn this {(args.Count() > 0 ? $"{args[0]} or" : "")} adder but my owner was too lazy. :(" }
```
jaymo1011, line 91: Wouldn't `args.Length` work here without the need of LINQ?

SaSiNO97:
> Why is the `OnClientResourceStart` event handler still registered manually instead of using `EventHandlerAttribute` or even in the code in the first place when it isn't used by the car spawner logic? It'd make more sense to have it be removed entirely in the context of a newbie friendly guide :)

Wait what :/ Because I didn't even know a `EventHandlerAttribute` even existed... This shows how seriously we need an updated documentation.

content/docs/scripting-manual/introduction/creating-your-first-script-csharp.md
SaSiNO97: Yeah true. I personally use LINQ a lot, that's why I used it there, however, it's just a personal preference and `string.Length` may be faster.

content/docs/scripting-manual/introduction/creating-your-first-script-csharp.md
```diff
@@ -138,31 +130,33 @@ RegisterCommand("car", new Action<int, List<object>, string>(async (source, args
...
+There may be C# wrappers here, but it's important to reify the use of natives as you will use them a lot when scripting. Make sure you have the `using static CitizenFX.Core.Native.API;` directive in your class.
 
 #### Step 2: Creating the vehicle
 Using the client side C# wrapper class `World`, we call the `CreateVehicle` method which takes a model, `Vector3` position, and `float` heading as arguments. This is the great thing about C#. You have access to a method supplied by _us_ such that you don't have to request and load a model like you would in Lua. This method returns us a `Vehicle` object. If you have experience with *ScriptHookV.NET* you may recognize these classes. The C# wrapper of FiveM is very similar.
 
+An important thing to notice is that the `CreateVehicle` method is [asnychronous](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/); this means that the wrapping method, in our case `CarCommand` must be `async` too, since we want to wait for our vehicle to be created before proceeding.
```
lze3, line 158: Typo

```suggestion
An important thing to notice is that the `CreateVehicle` method is [asynchronous](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/); this means that the wrapping method, in our case `CarCommand` must be `async` too, since we want to wait for our vehicle to be created before proceeding.
```

lze3 (changes requested):

PR #233, opened 2021-02-05
Added vehicle colours to game references.

The-Neco:
Added vehiclecolours.md to game references with images.

This is my first docs PR so please correct me if I have done something wrong :)


AvarianKnight:
Not the biggest fan of the split colors, given the age of the PR this can be fixed after the fact.

PR #234, opened 2021-02-07
Added documentation for chat exports.

The-Neco:
Added full documentation on chat exports.


4mmonium:
This documentation seems neat, I'll test it to make sure the examples work, just in case 😄

VIRUXE:
These changes are yet to be commited.

PR #238, opened 2021-02-12
feat(game-references): Add explosion types to the game references category.

spoty123:


tabarra:
This does not match with the ones from [txAdmin's codebase](https://github.com/tabarra/txAdmin/blob/master/scripts/sv_logger.js#L160).  
  
I know that I personally validated that list quite well when I added that code (Nov 24, 2019).  
So either this commit's list is wrong, or the one from txAdmin is and nobody complained ever.

spoty123:
> I know that I personally validated that list quite well when I added that code (Nov 24, 2019).

I made/validated that list from both "data dumps" and existing lists.
Also by testing, I can confirm with the 'explosionEvent' that a grenade explosion will cause id 0, grenade launcher will cause id 1 and so on.
The only other thing that is different is the fact that I added some new explosions that came with the 'Cayo Perico' Update.

> So either this commit's list is wrong, or the one from txAdmin is and nobody complained ever.

I personally just discovered that list even existed inside of txAdmin. 
Again, I don't believe that the list I committed is wrong since it corresponds with the 'explosionEvent'.

Let me know if you believe there is something I am missing here.

tabarra:
So in conclusion, the explosions ID change depending on game build?!  
Interesting, I hope someone tests this better, and I will be happy to update txAdmin as needed.

spoty123:
> So in conclusion, the explosions ID change depending on game build?!

I don't think so? Other than the fact that new explosions get added with almost every DLC.
(Which adds more ids, doesn't change the existing ones, as far as I know.)

PR #241, opened 2021-03-15
Use CommandAttribute instead of API.RegisterCommand

Sampleeeee:
I feel this is much easier to understand than creating a new Action, and what not. Also makes code look much cleaner.

### DIFF
--- a/content/docs/scripting-manual/introduction/creating-your-first-script-csharp.md
+++ b/content/docs/scripting-manual/introduction/creating-your-first-script-csharp.md
@@ -22,10 +22,7 @@ namespace MyResourceNameClient
 {
     public class Class1 : BaseScript
     {
-        public Class1()
-        {
-
-        }
+    
     }
 }
 ```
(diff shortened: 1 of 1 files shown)


content/docs/scripting-manual/introduction/creating-your-first-script-csharp.md
```diff
@@ -42,24 +39,15 @@ namespace MyResourceNameClient
 {
     public class Class1 : BaseScript
     {
-        public Class1()
-        {
-            EventHandlers["onClientResourceStart"] += new Action<string>(OnClientResourceStart);
```
4mmonium, line 47: Looks like you removed the class constructor and the event handler for `OnClientResourceStart`?

4mmonium:
Not entirely sure what happened here, it seems like you modified `creating-your-first-script-csharp` to process scripting commands, but the guide is explicitly written, _to explain a user how to go about creating their first script_. It should provide some basic examples, it looks like some of those examples were removed by looking at this PR, that should be corrected :-)

AvarianKnight:
This is already apart of the tracker https://github.com/citizenfx/fivem-docs/issues/466, going to close this as it will get rewritten later

PR #245, opened 2021-05-12
Update and rename compiling-fivem.md to building-fivem.md

aiden-z-source:


tabarra:
When you update AND rename, it gets harder to see what you changed since it's considered a new file.

aiden-z-source:
> 
> 
> When you update AND rename, it gets harder to see what you changed since it's considered a new file.

Yeah I didn't realize that at the time, however I do believe I did most of the changes on [09dfd18](https://github.com/citizenfx/fivem-docs/pull/245/commits/09dfd1881e729fa3807c23dd8538743111edcebf)

PR #253, opened 2021-06-07
Update txAdmin/_index.md

Konijntje-x:
Added some lines about starting txAdmin.


technetium-cfx:
This is already explained in the server installation guide.

PR #255, opened 2021-07-05
tweak(client-manual): add note about resmon use

flerovium-ak:
Adds a note about using `resmon` now that it needs Canary or `+set moo 31337`. Chose this location since it is the first result of "fivem resmon" on Google, but it might not be the best placement in the docs. I'm happy to relocate it if anyone has any better suggestions!

Either way: this should hopefully add some visibility on how to use `resmon` so people don't use Canary for no reason other than accessing `resmon`.

For reference, Google:
CleanShot 2021-07-05 at 21 41 22@2x


technetium-cfx:
This actually counts for more commands, and I suggest we take a more generic explanation as such. Perhaps this warrants a more clear flag name as well, but that's another topic.

MichaelCoding25:
@flerovium-ak This can be closes I believe because it was solved in this commit https://github.com/citizenfx/fivem-docs/commit/92d933da86887a4678d67d9195fd3dfa86d4523b

Issue #256, opened 2021-07-06
Update RegisterNetEvent Documentation


MichaelCoding25:
Since this commit ([citizenfx/fivem@81caf7b](https://github.com/citizenfx/fivem/commit/81caf7b38f0d5404f26edd75b71ba5d2987ffad2)) in January, you can also add a callback (function) in the RegisterNetEvent instead of having to also add AddEventHandler.

This is currently not documented and I think would be very beneficial if someone did.
I would do it myself but not sure exactly how to word it.

MichaelCoding25:
This was done by @The-Neco so I will be closing this now since it's done.

PR #257, opened 2021-07-06
tweak(lua/functions): Add optional arguments to RegisterNetEvent.

The-Neco:
Added the optional `callback` argument to RegisterNetEvent as pointed out by @MichaelCoding25 in issue #256.

### DIFF
--- a/content/docs/scripting-reference/runtimes/lua/functions/RegisterNetEvent.md
+++ b/content/docs/scripting-reference/runtimes/lua/functions/RegisterNetEvent.md
@@ -5,20 +5,31 @@ title: RegisterNetEvent
 <!-- alias: RegisterServerEvent for server only -->
 
 Marks the event safe for network use. Aka, allows you to trigger the eventName event on the client, from a server side
-script. Use [AddEventHandler][] to listen for the event after registering it.
+script. If you do not provide a callback function use [AddEventHandler][] to listen for the event after registering it.
 
 Syntax
 ------
 
 ```lua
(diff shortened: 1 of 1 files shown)


content/docs/scripting-reference/runtimes/lua/functions/RegisterNetEvent.md
```diff
@@ -5,21 +5,32 @@ title: RegisterNetEvent
 <!-- alias: RegisterServerEvent for server only -->
 
 Marks the event safe for network use. Aka, allows you to trigger the eventName event on the client, from a server side
-script. Use [AddEventHandler][] to listen for the event after registering it.
+script. If you do not provide a callback function use [AddEventHandler][] to listen for the event after registering it.
```
technetium-cfx, line 8: Might want to fix the link while we're at it 😛

technetium-cfx (changes requested):

content/docs/scripting-reference/runtimes/lua/functions/RegisterNetEvent.md
The-Neco: Does defining the link at the bottom not work on the docs page? That's how it was previously so I didn't change it, I will however if defining it in-line is better practice. (https://docs.fivem.net/docs/scripting-reference/runtimes/lua/functions/RegisterNetEvent/)

technetium-cfx: You are right indeed!

PR #265, opened 2021-08-15
update(resource-manifest): Add convar_category to resource-manifest.md.

The-Neco:
This commit adds information and examples on the `convar_category` manifest entry. This may be a lot of information for this section so the "types" may need to be moved to a separate section.

Example provided by @CodeWithBryan

### DIFF
--- a/content/docs/scripting-reference/resource-manifest/resource-manifest.md
+++ b/content/docs/scripting-reference/resource-manifest/resource-manifest.md
@@ -288,6 +288,58 @@ When present, disables the custom C# task scheduler on the server. This will inc
 clr_disable_task_scheduler 'yes'
 ```
 
+### convar_category
+
+When present adds the specified convars to the 'Project Settings' page in FXDK.
+
+#### Convar types:
+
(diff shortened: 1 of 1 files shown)


content/docs/scripting-reference/resource-manifest/resource-manifest.md
```diff
@@ -288,6 +288,58 @@ When present, disables the custom C# task scheduler on the server. This will inc
 clr_disable_task_scheduler 'yes'
 ```
 
+### convar_category
```
4mmonium, line 291: Looks like this was added later on as part of the FxDK docs, but the effort is appreciated nonetheless!

https://github.com/citizenfx/fivem-docs/commit/bfec0b2c866a5d1c6b4284869af20217371b07d9#diff-311e0022c022ac7ffe009570c09040c718065154622cc6b9daabb1a3bd1d3810R40

4mmonium (approved):
Upon further consideration, it looks like it would be great to have this merged for visibility, thank you for your contribution!

PR #267, opened 2021-08-20
Add a new common issue

MasiBall:
Add 'server doesn't show up on the server list' to common issues. Hopefully reduces the amount of people asking 'why my server isn't showing on the server list'.


tabarra:
Might be worth pointing out that the keymaster has a great debug message to help find out the reason for it not to show.

MasiBall:
This is not needed anymore as the issue has already been fixed.

PR #276, opened 2021-10-03
Simplify readability with all-in-one RegisterNetEvent

Ekinoxx0:
Hi there,

with the addition of the compact RegisterNetEvent that also do AddEventHandler in the background,
do we want to suggest new dev to use it in place of AddEventHandler + RegisterNetEvent ?

That would simplify readability of future resources, but also hide the little "complexity" hided behind this.

### DIFF
--- a/content/docs/scripting-manual/working-with-events/listening-for-events.md
+++ b/content/docs/scripting-manual/working-with-events/listening-for-events.md
@@ -59,10 +59,7 @@ To register an event as a network event, place this in your client/server script
 
 **Lua**
 ```lua
-RegisterNetEvent("eventName")
-
--- The event handler function follows after registering the event first.
-AddEventHandler("eventName", function(eventParam1, eventParam2)
+RegisterNetEvent("eventName", function(eventParam1, eventParam2)
     -- Code here will be executed once the event is triggered.
 end)
 ```
--- a/content/docs/scripting-reference/runtimes/lua/functions/TriggerClientEvent.md
+++ b/content/docs/scripting-reference/runtimes/lua/functions/TriggerClientEvent.md
@@ -22,13 +22,27 @@ Examples
 --------
 
 -- CLIENT
+
+Lua's traditional way to register a net event and add an event handler:
 ```lua
-RegisterNetEvent('eventName', function(text)
-  print(('I just received %s from the server'):format(text)) -- I just received Hello world! from the server
+RegisterNetEvent("eventName")
+-- The event handler function follows after registering the event first.
+AddEventHandler("eventName", function(eventParam1, eventParam2)
+    -- Code here will be executed once the event is triggered.
 end)
 ```
+
+Lua's new simplified way:
+```lua
+RegisterNetEvent("eventName", function(eventParam1, eventParam2)
+    -- Code here will be executed once the event is triggered.
+end)
+```
+
 -- SERVER
 ```lua
 TriggerClientEvent('eventName', playerId, 'Hello world!')
 ```
+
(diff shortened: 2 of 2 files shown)


4mmonium:
Looks like this is already done here. Thanks for the effort though!

4mmonium:
I think it would be best, if you suggested this as an improvement rather than as a way to replace old functionality _that is still there_. Many people do probably still depend on the old one, so I think the best way to present this would be in the following fashion :-)

**Lua's traditional way to register a net event and add a handler:**
```lua
RegisterNetEvent("eventName")
-- The event handler function follows after registering the event first.
AddEventHandler("eventName", function(eventParam1, eventParam2)
    -- Code here will be executed once the event is triggered.
end)
```

**Lua's new simplified way:**
```lua
RegisterNetEvent("eventName", function(eventParam1, eventParam2)
    -- Code here will be executed once the event is triggered.
end)
```

That way both ways are valid.

Ekinoxx0:
Thanks for answering :)

Your suggestion is great so I added it to this PR, you're right about also suggesting the old way with an explanation

4mmonium:
Looks good to me!
Thank you for the prompt update, once you feel your PR is ready, make sure to squash it all into one commit, this is done to avoid redundant commits within the repository. If you're not sure on how to squash your commits, I suggest you to read [this](https://stackoverflow.com/a/5189600) answer from stackoverflow or [this](https://www.git-tower.com/learn/git/faq/git-squash).

In your case you would most likely have to squash your last two commits, except for the last one so: `git rebase -i HEAD~3`
Most likely a `vi` text editor will show in the console, press `insert` on your keyboard to go into 'insert' mode. From there, select the ones you want to squash and 'pick' a parent commit.

After squashing everything, you would most likely have to push to the repository again, so use `git push --force` to force push your changes and overwrite your previous commits. Please note that it's not advised to use `git push --force` when working with other team-mates and the intention is overwriting changes, in that case you may `git push --force-with-lease` to see if your changes wouldn't overwrite any other recent changes your team-mates did. You can check the difference between both [here](https://stackoverflow.com/a/52823955).

Be very careful when doing anything with git, you wouldn't want to lose the work you have done! 🙂

content/docs/scripting-manual/working-with-events/listening-for-events.md
```diff
@@ -59,10 +59,7 @@ To register an event as a network event, place this in your client/server script
 
 **Lua**
 ```lua
-RegisterNetEvent("eventName")
-
--- The event handler function follows after registering the event first.
-AddEventHandler("eventName", function(eventParam1, eventParam2)
+RegisterNetEvent("eventName", function(eventParam1, eventParam2)
```
4mmonium, line 62: If you could add back the traditional way to do it here as well, it would be great, thanks!

content/docs/scripting-reference/runtimes/lua/functions/TriggerClientEvent.md
```diff
@@ -23,14 +23,26 @@ Examples
...
-AddEventHandler('eventName', function(text)
-  print(('I just received %s from the server'):format(text)) -- I just received Hello world! from the server
+RegisterNetEvent("eventName")
+-- The event handler function follows after registering the event first.
+AddEventHandler("eventName", function(eventParam1, eventParam2)
+    -- Code here will be executed once the event is triggered.
 end)
 ```
+
+Lua's new simplified way:
```
AvarianKnight, line 35:
```suggestion
Lua's simplified way:
```

PR #280, opened 2021-11-08
Add nuget package reference for C# Client

SQKo:
Add a link to [official CitizenFX NuGet Package](https://www.nuget.org/packages/CitizenFX.Core.Client/) for the **CitizenFX.Core** client script dependency, useful for those who develop on separate machine where FiveM is not installed


SQKo:
I forgot the up to date package is only the CitizenFX.Core.Client not CitizenFX.Core

PR #281, opened 2021-11-08
Update creating-your-first-script-csharp.md

SQKo:
Add a client-specific `ClientScript` and server-specific `ServerScript` class as alternative to `BaseScript`, which was added since artifact 3395 https://github.com/citizenfx/fivem/commit/304f787be3ae9a2a6f0365688318512ea69b1cfe


4mmonium (commented):
This looks good to me, sorry for the super late response. Perhaps you could provide some usage example? :-)

SQKo:
There is no example actually, unless you want me to replace "BaseScript" with "ClientScript"/"ServerScript" in examples of another pages

As far I know this was added only to prevent possible ambiguity/conflict when building and linking the .dll

AvarianKnight:
This should probably be explained better within the example, given the age of this PR I'm going to close this and add it to the tracking issue https://github.com/citizenfx/fivem-docs/issues/466

PR #282, opened 2021-11-14
Info about Node 16.x

Mojito-Fivem:
Includes details that server builds greater than 4800 run node 16.x


tabarra:
Well, it was not a single commit, but I'm pretty sure if we had to draw a line it would be closer to 4548 than 4800.  
  
https://github.com/citizenfx/fivem/commit/98e9c7bb8a0161707de50196e63d46065d915e7d

blattersturm:
I don't think it makes sense to actually define version ranges in the docs, as otherwise we'd also have to mention 8.x somewhere.

PR #283, opened 2021-11-18
Fix for broken installation guide list

GJSBRT:
Fix for broken installation guide list on https://docs.fivem.net/docs/server-manual/setting-up-a-server/


roentgenium-cfx:
See #299.

PR #289, opened 2022-01-07
Add FiveM identifier to `GetPlayerIdentifiers`

FlawwsX:


blattersturm:
I'd rather see this weird example rewritten, also as even with this change it still does not 'show all possible identifiers'.

FlawwsX:
```lua
RegisterCommand('identifiers', function(source)
    local identifiers = {}
    for k,v in pairs(GetPlayerIdentifiers(source)) do
        local colonPosition = string.find(v, ":") - 1
        local identifierType = string.sub(v, 1, colonPosition)
        identifiers[identifierType] = v
    end
    -- If I wanted to get their fivem identifier, I would just do:
    print(identifiers["fivem"])
end, false)
```
Would something like this work?
May not be exactly what you wanted but it does show all possible identifiers and the code is easier to understand.

AvarianKnight:
That probably would be a better example, int should probably be included that if players only need a specific identifier they can use https://docs.fivem.net/natives/?_0xA61C8FC6

PR #299, opened 2022-02-14
Update setting up a server guide.

4mmonium:
I mainly updated the guide so it's more organized and added a zap hosting section to it. For the windows guide I cleared up on how to clone a git repo to a directory since I'm sure users that are new to `git` won't know this even though they can Google.

### DIFF
--- a/content/docs/_index.md
+++ b/content/docs/_index.md
@@ -1,5 +1,5 @@
 ---
-title: Home
+title: 👋🏼 Home
 layout: single
 linkTitle: "Documentation"
 menu:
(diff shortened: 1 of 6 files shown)


blattersturm:
If the Windows guide still mentions the VC redist, this ain't needed anymore.

4mmonium:
> If the Windows guide still mentions the VC redist, this isn't needed anymore.
> 
> I'll push a draft PR soon that had some local changes to this and other pages.

That's great, I will remove that if it's unneeded and will merge your drafted changes if that's ok 😀

PR #301, opened 2022-02-20
Document how to use Lua 5.4 and power patches

4mmonium:
This commit adds documentation on how to use Lua 5.4 and documents the
power patches provided by this new release. Most of the documentation
for this comes from the readme (https://github.com/gottfriedleibniz/lua/blob/luaglm/README.md),
'as is' and the lua-users Wiki: http://lua-users.org/wiki/LuaPowerPatches#LUA_54_patches
I tried to not modify any of the documentation, only adding certain
markdown styling and some extra explanation to specific terms which may not be
known to new users.

### DIFF
--- a/content/docs/scripting-reference/runtimes/lua/power-patches.md
+++ b/content/docs/scripting-reference/runtimes/lua/power-patches.md
@@ -1,0 +1,286 @@
+---
+title: Lua power patches
+weight: 516
+---
+
+# Lua Power Patches
+
+# Lua 5.4
+
+
+## Compound Operators
+
+Add ``+=, -=, *=, /=, <<=, >>=, &=, |=, and ^=`` to the language. The increment
+and decrement operators (``++, --``) have not been implemented due to one of
+those operators being reserved.
+
+## Safe Navigation
+
+An indexing operation that suppresses errors on accesses into undefined tables
+(similar to the safe-navigation operators in C#, Kotlin, etc.), e.g.,
+
+```lua
+t?.x?.y == nil
+```
+
+## In Unpacking
+
+Support for unpacking named values from tables using the ``in`` keyword, e.g,
+
+```lua
+local a,b,c in t
+```
+
+is functionally equivalent to:
+
+```lua
+local a,b,c = t.a,t.b,t.c
+```
+
+## Set Constructors
+
+Syntactic sugar to improve the syntax for specifying sets, e.g.,
+
+```lua
+t = { .a, .b }
+```
+
+is functionally equivalent to:
+
+```lua
+t = { a = true, b = true }
+```
+
+## C-Style Comments
+
+Support for C-style block comments: ``/* Comment */``, e.g.,
+
+```lua
+print("Hello, World!") /* Comment */
+```
+
+## Compile Time Jenkins' Hashes
+
+String literals wrapped in back-ticks are Jenkins' one-at-a-time hashed when
+parsed.
+
+```lua
+> `Hello, World!`
+1395890823
+```
+
+For runtime hashing, the `joaat` function is included in the base library:
+
+```lua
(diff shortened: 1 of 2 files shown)


4mmonium:
> Currently, FiveM is compiled with a [different tree](https://github.com/citizenfx/lua/tree/2927a4be425e85b911764c90636c4c67fed9e640) of LuaGLM that does not include all of the power patches [gottfriedleibniz/lua:luaglm](https://github.com/gottfriedleibniz/lua/tree/luaglm) has and not _all_ of them are [enabled](https://github.com/citizenfx/fivem/blob/d1a48fc1af5b16650a7b60a3c840114da7dbe033/code/vendor/lua54.lua#L41) either :)

Thanks, I guess I will review this and remove any power patches that don't apply from the docs 😃

content/docs/scripting-reference/runtimes/lua/power-patches.md
```diff
@@ -0,0 +1,286 @@
...
+local a,b,c in t
+```
+
+is functionally equivalent to:
+
+```lua
+local a,b,c = t.a,t.b,t.c
+```
+
+## Set Constructors
```
4mmonium, line 40: Remove, doesn't exist in defines.

content/docs/scripting-reference/runtimes/lua/power-patches.md
```diff
@@ -0,0 +1,286 @@
...
+```
+
+```lua
+> joaat("CPed")
+2491553369
+```
+
+## Short Function Notation or Shorthand lambdas (5.4.2)
+
+Syntactic sugar for writing concise anonymous functions of the form `|a, b,
```
4mmonium, line 89: Removing, couldn't find a reference to this

content/docs/scripting-reference/runtimes/lua/power-patches.md
jaymo1011: Set constructors do exist!
https://github.com/citizenfx/fivem/blob/92cbb7abb99312b114c532475455a6526241481f/code/vendor/lua54.lua#L44

4mmonium: Why isn't it referenced anywhere in the source code though? I find this weird, if it's a define, it should be somewhere, no?
https://github.com/search?q=GRIT_POWER_TABINIT&type=code

jaymo1011: GitHub doesn't index forks for code searching, both citizenfx/lua and gottfriedleibniz/lua just so happen to be forks.

4mmonium: Ah, thanks! Will download the source code and try and find locally then

4mmonium:
Ah, this is also super useful, just found thanks to local search, completely overlooked this 😀
https://github.com/citizenfx/lua/blob/2927a4be425e85b911764c90636c4c67fed9e640/CMakeLists.txt#L454

content/docs/scripting-reference/runtimes/lua/power-patches.md
```diff
@@ -0,0 +1,286 @@
...
+--
+-- Finally, tables with 'protected' metatables, i.e., a '__metatable' field,
+-- cannot be frozen.
+t = table.freeze(t)
+
+-- Return true if the provided table is configured as readonly; false otherwise.
+result = table.isfrozen(t)
+```
+
+## Save/restore readline history (5.4.2)
```
4mmonium, line 256: This and Advanced readline support (5.4, 5.3, 5.2, 5.1, 5.0) aren't enabled, `GRIT_POWER_READLINE_HISTORY` isn't set in the defines.

PR #322, opened 2022-04-26
Fix default variables for sv_authMinTrust and sv_authMaxVariance

FlawwsX:
image

According to the above, these are documented, on the docs, the wrong way around.


4mmonium:
Looks good to me, based on this:
https://github.com/citizenfx/fivem/blob/a19789cd3f5029011993a32fe8ed46f8787e447c/code/components/citizen-server-impl/src/InitConnectMethod.cpp#L301

PR #338, opened 2022-07-01
Add C# Server Side Documentation

Luke20201:
I saw C# only had client-side Documentation, so I decided to spend a few hours documenting what I felt was the most important and useful for FiveM developers in C#.

This covers;

- Setting up the environment (using NuGet)
- Hooking events and using [FromSource]
- Server Callbacks
- MySQL

I think I explained everything in good detail, though this doc _assumes_ the user read the client-side documentation first.

### DIFF
--- a/content/docs/scripting-manual/introduction/csharp.md
+++ b/content/docs/scripting-manual/introduction/csharp.md
@@ -63,8 +63,8 @@ This sounds much more difficult than it actually is, but it's really not hard!
    Ok, there's quite a bit in here. So let me explain how events are hooked.
    Events are triggered either by a script or by the server when certain conditions are met or triggered. These events come with special parameters, like this one! So let's explain the parameters!
   1. `Player` is a special class that has a massive amount of information about an individual user. We are going to be using this a lot.
-  2. `string` is going to contain the name of the player in **this** instance.
-  3. `dynamic`'s are two very special types we will be working with here. The first dynamic contains a reason for kicking the player, and the second is our deferral. 
+  2. `string` is going to contain the name of the player and a reason for kicking them(if applicable!) in **this** instance.
+  3. `dynamic` is a special type we will be working with here. The dynamic here is our deferral that we will use to delay the players connection if necessary.. 
   It's important to explain what a dynamic is, before we proceed. Dynamics can be pretty much anything, this is because the type of the dynamic is only checked at runtime, not by your compiler.
   ```csharp
   dynamic number = 123; // This can be a dynamic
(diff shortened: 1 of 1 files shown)


blattersturm:
While we appreciate the effort, this seems to be full of outdated/weird/wrong information.

The C# docs already mention setting up a server script because the template does so already.

(I am currently not able to provide more detailed comments as I'm on my phone, just remarking to other project members that this should not be merged!)

Luke20201:
Hi,

I was 100% confused on whether to use NetworkCallbackDelegate or not, I saw a lot of people saying it didn't work, and a few others saying that CFX confirmed it working. I went with it because I have had no issues using it, but if there is an issue with it I understand.

Could you elaborate on anything else I may have confused so it can be corrected?

thorium-cfx:
As an extra to the rewriting: don't use `dynamic` types unless you really need to make use of the dynamic-ness it offers.

`dynamic` is compiled to type `object` with a lot of extra IL code to make sure it can do all the stuff you do with it. This also makes JITting the code slower (I've seen freezes up to several seconds with some specific code). Better to use `var` (or the type) for local variable declarations and the exact type as parameters. Also, although small, `object` also has a (hidden) runtime `pack`/`unpack` cost. Better not teaching new programmers bad habits ;)

Luke20201:
> As an extra to the rewriting: don't use `dynamic` types unless you really need to make use of the dynamic-ness it offers.
> 
> `dynamic` is compiled to type `object` with a lot of extra IL code to make sure it can do all the stuff you do with it. This also makes JITting the code slower (I've seen freezes up to several seconds with some specific code). Better to use `var` (or the type) for local variable declarations and the exact type as parameters. Also, although small, `object` also has a (hidden) runtime `pack`/`unpack` cost. Better not teaching new programmers bad habits ;)

Hi thorium,

The reason for using dynamic was that was how it is done in the [playerConnecting event](https://docs.fivem.net/docs/scripting-reference/events/list/playerConnecting/) within the scripting reference. Though I did assume setKickReason could've been a string, I just didn't test it to confirm. I'll edit that shortly.

thorium-cfx:
> The reason for using dynamic was that was how it is done in the [playerConnecting event](https://docs.fivem.net/docs/scripting-reference/events/list/playerConnecting/) within the scripting reference. Though I did assume setKickReason could've been a string, I just didn't test it to confirm. I'll edit that shortly.

I see :thinking: 
From looking into the source it looks like that parameter will be of type `CallbackDelegate` (`CitizenFX.Core.CallbackDelegate` to be precise). The deferrals on the other hand might just be better of as a `dynamic`, looks like that would be an `ExpandoObject` which is related to them.

PavlovTM:
Hello, I am just reading through this post. Since I am also interested in improving the docs.

Only unfortunately I notice that it is always mentioned how some elements like "NetworkCallBackDelegate" are deprecated, but no working alternative is mentioned.

Is there any update?

AvarianKnight:
Going to close this, C# documentation will be improved later.

PR #340, opened 2022-07-23
FiveM on Linux 

Rosentti:
DXVK has added support for shared resources, and the client now works on Linux without the anticheat.

### DIFF
--- a/content/docs/support/client-faq.md
+++ b/content/docs/support/client-faq.md
@@ -11,8 +11,9 @@ No. FiveM is developed exclusively for the PC version of GTA V. There are no pla
 Will FiveM run on Linux?
 ------------------------
 
-You can run the FiveM client on Linux with DXVK, however the anticheat is not supported so you will not be able to play on most servers. You can however make your own server, just remember to remove svadhesive so you can play on Linux. 
+You can run the FiveM client on Linux, although the anticheat is not supported so you will not be able to play on most servers. You can make your own server from which you may remove svadhesive if you want Linux users to be able to join. This makes the anticheat not required for players that join, so tread carefully!
 Running the game on Linux is not officially supported, and it may be buggy.
+Running servers on Linux is supported, anticheat included. 
 
 Where is FiveM installed?
 -------------------------
--- a/content/docs/support/client-faq.md
+++ b/content/docs/support/client-faq.md
@@ -11,12 +11,9 @@ No. FiveM is developed exclusively for the PC version of GTA V. There are no pla
 Will FiveM run on Linux?
 ------------------------
 
-You can run the FiveM client on Linux through Wine/Proton, although the anticheat is not supported so you will not be able to play on most servers. It's not currently planned to make the anticheat support Linux.
+You can run the FiveM client on Linux through Wine/Proton, but the anti-cheat will not work, meaning you will only be able to join servers that have disabled `svadhesive`.
 
-You can make your own server from which you may remove svadhesive if you want Linux users to be able to join. This makes the anticheat not required for players that join, so tread carefully!
-
-Running the game on Linux is not officially supported, and it may be buggy and break with updates.
-Running servers on Linux is supported, anticheat included. 
+**NOTE**: Running on Linux **might** work, but there is no plans to anti-cheat support. Issues or topics pertaining to this will likely be closed.
 
 Where is FiveM installed?
 -------------------------
--- a/content/docs/support/client-faq.md
+++ b/content/docs/support/client-faq.md
@@ -11,7 +11,8 @@ No. FiveM is developed exclusively for the PC version of GTA V. There are no pla
 Will FiveM run on Linux?
 ------------------------
 
-The FiveM client cannot run on Linux due to [no support](https://github.com/doitsujin/dxvk/issues/899) for shared resources in DXVK. You can run a FiveM server on Linux.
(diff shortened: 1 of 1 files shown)


tommy141x (approved):

content/docs/support/client-faq.md
```diff
@@ -11,7 +11,8 @@ No. FiveM is developed exclusively for the PC version of GTA V. There are no pla
 Will FiveM run on Linux?
 ------------------------
 
-The FiveM client cannot run on Linux due to [no support](https://github.com/doitsujin/dxvk/issues/899) for shared resources in DXVK. You can run a FiveM server on Linux.
+You can run the FiveM client on Linux with DXVK, however the anticheat is not supported so you will not be able to play on most servers. You can however make your own server, just remember to remove svadhesive so you can play on Linux.
```
PsychoShock, line 14: I don't think having docs saying that is useful. 

Also note that adding "the anticheat is not supported" is not really recommanded.

Rosentti: I've reworded the section. Please re-review.

What exactly isn't useful?

tommy141x:
Not sure I understand why you gave this PR a thumbs down and stated that it's not useful. Claiming that FiveM cannot run on Linux due to DXVK with a linked issue that was resolved almost 5 years ago is not useful. Documentation is meant to be informative, I suggest either update the documentation or just remove the entire bit.

PsychoShock:
I don't recommand the merge of this PR for multilples reasons. This is not the place to debate. I'm just giving my personnal opinion. At the end of the day, only @4mmonium will be able to review it.

Rosentti:
Aren't pull requests the best place to debate?
After all, PRs are the way software on Github changes due to community contributions. 

Also, claiming DXVK doesn't support shared resources while it actually does is detrimental to DXVK, so this should at the very least be addressed and made clear that DXVK's shared resources aren't the problem. 

If you're against the merge for "multiple reasons", you should probably list them.

AvarianKnight:
I don't think there is a good reason to not merge this (sorry for the double negative).

This doc was written years ago and the DXVK is no longer the limitation.

I don't think that the wording should be so passive with the wording though.

> You can run the FiveM client on Linux through Wine/Proton, although the anticheat is not supported so you will not be able to play on most servers. It's not currently planned to make the anticheat support Linux.

Should probably be 

> You can run the FiveM client on Linux through Wine/Proton, but the anticheat will not work on Linux so you will not be able to play on servers requiring authentication.
> **NOTE:** Although running FiveM on Linux works, it is not supported, any issues pertaining to Linux are likely to be ignored and closed.

It doesn't make sense to include 

> Running servers on Linux is supported, anticheat included. 

As this section is the client faq

GuiTheDev (approved):

sekwah41:
Agreed, with the area for debate. If a PR discussion isn't the area for debate as to whether a PR is valid, then where is?

Also its not fair to blame another project for the incompatibility when the incompatibility is solely on the Anticheat at this point. Changing the wording also makes it clear its possible to run without the anticheat on.

If you have reasons not to want to merge it, it only makes sense to share it so that the issues can be addressed to improve the PR or justify why it would be rejected.

content/docs/support/client-faq.md
```diff
@@ -11,7 +11,12 @@ No. FiveM is developed exclusively for the PC version of GTA V. There are no pla
 Will FiveM run on Linux?
 ------------------------
 
-The FiveM client cannot run on Linux due to [no support](https://github.com/doitsujin/dxvk/issues/899) for shared resources in DXVK. You can run a FiveM server on Linux.
+You can run the FiveM client on Linux through Wine/Proton, although the anticheat is not supported so you will not be able to play on most servers. It's not currently planned to make the anticheat support Linux.
+
+You can make your own server from which you may remove svadhesive if you want Linux users to be able to join. This makes the anticheat not required for players that join, so tread carefully!
+
+Running the game on Linux is not officially supported, and it may be buggy and break with updates.
+Running servers on Linux is supported, anticheat included.
```
AvarianKnight, line 19:
```suggestion
You can run the FiveM client on Linux through Wine/Proton, but the anticheat will not work so you will not be able to play on servers requiring authentication.

**NOTE**: Although running FiveM on Linux may work, it is not supported, any issues pertaining to Linux are likely to be  closed.
```

content/docs/support/client-faq.md
```diff
@@ -11,7 +11,9 @@ No. FiveM is developed exclusively for the PC version of GTA V. There are no pla
 Will FiveM run on Linux?
 ------------------------
 
-The FiveM client cannot run on Linux due to [no support](https://github.com/doitsujin/dxvk/issues/899) for shared resources in DXVK. You can run a FiveM server on Linux.
+You can run the FiveM client on Linux through Wine/Proton, but the anti-cheat will not work, meaning you will only be able to join servers that have disabled `svadhesive`.
+
+**NOTE**: Running on Linux **might** work, but there is no plans to anti-cheat support. Issues or topics pertaining to this will likely be closed.
```
AvarianKnight, line 16:
```suggestion
**NOTE**: Running on Linux **might** work, but there is no plans for anti-cheat support. Issues or topics pertaining to this will likely be closed.
```

AvarianKnight (commented):
I made a typo

PR #343, opened 2022-08-04
Add voice guide (open for discussion)

4mmonium:
I mainly created this guide upon request, and as a way to spark up discussion and share research.
Critics are welcome, if something seems off, do please let me know and I will do my best to come up with a solution, thanks!

### DIFF
--- a/content/docs/scripting-manual/voice/_index.md
+++ b/content/docs/scripting-manual/voice/_index.md
@@ -1,0 +1,260 @@
+---
+title: Voice
+weight: 450
+---
+
+FiveM has been using the [mumble](https://www.mumble.info/) client to handle in-game voice communication for a while now, below we will explain how you can use certain mumble natives, including a couple guides to handle submixes and NativeAudio.
+
+# Console Variables
+
+These variables can be enabled/disabled on the client by pressing `F8` while FiveM is running.
+
+### `voice_inBitrate [bitrate]`
+
+Allows you to set the voice [bitrate](https://en.wikipedia.org/wiki/Bit_rate), ranges from `16000` to `128000`. Default is `48000`. The greater the bitrate, the better the audio quality.
+
+### `voice_use2dAudio [true/false]`
+
+Uses 2D Audio exclusively. This is set to false by default.
+
+### `voice_use3dAudio [true/false]`
+
+Considered deprecated on FiveM, but available on RedM.  
+Uses 3D (directional) Audio exclusively. This is set to false by default.
+Currently, directional audio's position is relative to the game camera, a solution is being worked on so directional audio is relative to the player's ped entity instead.
+
+### `voice_useSendingRangeOnly [true/false]`
+
+A Convar that can be used to only hear other clients within their range. This is set to false by default.
+
+### `voice_useNativeAudio [true/false]`
+
+Enables the game's native audio including filter support. This should be enabled if you plan to use submixes. This is set to false by default.
+
+# Talker Proximity
+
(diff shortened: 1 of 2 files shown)


content/docs/scripting-manual/voice/_index.md
```diff
@@ -0,0 +1,258 @@
...
+
+### `voice_inBitrate [bitrate]`
+
+Allows you to set the voice [bitrate](https://en.wikipedia.org/wiki/Bit_rate), ranges from `16000` to `128000`. Default is `48000`. The greater the bitrate, the better the audio quality.
+
+### `voice_use2dAudio [true/false]`
+
+Uses 2D Audio exclusively. This is set to false by default.
+
+### `voice_use3dAudio [true/false]`
```
AvarianKnight, line 20: It should probably be noted that `voice_use3dAudio` is considered deprecated for FiveM and should only really be used for RedM

content/docs/scripting-manual/voice/_index.md
```diff
@@ -0,0 +1,258 @@
...
+
+A Convar that can be used to only hear other clients within their range. This is set to false by default.
+
+### `voice_useNativeAudio [true/false]`
+
+Enables the game's native audio including filter support. This should be enabled if you plan to use submixes. This is set to false by default.
+
+# Talker Proximity
+
+You can use [`MUMBLE_SET_TALKER_PROXIMITY`](https://docs.fivem.net/natives/?_0x74E927B0) as well as [`NETWORK_SET_TALKER_PROXIMITY`](https://docs.fivem.net/natives/?_0xCBF12D65F95AD686) (they both pretty much do the same thing, only difference is that `NETWORK_SET_TALKER_PROXIMITY` also [calls](https://github.com/citizenfx/fivem/blob/928bd19f886d66d3f85340e65df69ae97ab91dc7/code/components/gta-net-five/src/MumbleVoice.cpp#L1343) the original game method) to limit distance between players when communicating.
```
AvarianKnight, line 35: `NETWORK_SET_TALKER_PROXIMITY` is also not a thing in RedM, so it might be a good idea to prefer  `MUMBLE_SET_TALKER_PROXIMITY` over the GTA specific native

content/docs/scripting-manual/voice/_index.md
```diff
@@ -0,0 +1,258 @@
...
+end, false)
+```
+
+### Writing the code (client-side)
+
+We will first register an event named `onPlayerChangeVoiceChannels`, we will be using this event to iterate through the list of clients we previously mentioned (which the server is sending us) and set their volume. We will also be setting our voice channel through here by calling [`MUMBLE_SET_VOICE_CHANNEL`](https://docs.fivem.net/natives/?_0x8737EEE8).
+
+```lua
+RegisterNetEvent("onPlayerChangeVoiceChannels")
+AddEventHandler("onPlayerChangeVoiceChannels", function(clients, channel, state)
```
AvarianKnight, line 217: Nitpick here, why not just use `RegisterNetEvent("onPlayerChangeVoiceChannels", function(clients, channel, state)`

AvarianKnight (changes requested):

content/docs/scripting-manual/voice/_index.md
4mmonium: Ah, will do! I did this on FiveM hence why. Thanks for jumping in, was waiting for you, didn't wanna mention/ping 😄

content/docs/scripting-manual/voice/_index.md
4mmonium: Will do, seems like I went the legacy way of doing events here.

PR #349, opened 2022-09-15
Added a new Game Event

EeyZed:
Added CEventNetworkPlayerEnteredVehicle Event


ThePawlow:
This is confusing, does the Event exist already?
If so, where is the Leave Event?

DonHulieo:
There is no leave event, unfortunately, here's the event triggered and printed to console.

image

PR #355, opened 2022-10-26
Correct typos throughout docs wiki.

4mmonium:
Thanks to gottfriedleibniz and others for finding them.


4mmonium:
> fact-sheet

Wow, nice job, will add these too

PR #372, opened 2023-02-03
fix(scripting-reference): vector4 docs typo

goncalobsccosta:


content/docs/scripting-reference/runtimes/lua/functions/vector4.md
```diff
@@ -65,8 +65,8 @@ local v = vector4(1, 2, 3, 4)
 print(#v) -- prints `5.477`
 
 -- Vectors can be normalized:
-local v = vector3(1, 2, 3, 4)
-print(norm(v)) -- prints `vector3(0.18, 0.37, 0.55, 0.73)`
+local v = vector4(1, 2, 3, 4)
+print(norm(v)) -- prints `vector4(0.18, 0.37, 0.55, 0.73)`
```
4mmonium, line 69: Looks good, this below has different output, but if we want to be exact:

```lua
print(norm(v)) -- prints `vector4(0.1825742, 0.3651484, 0.5477225, 0.7302967)`
```

Just a nitpick really, nice work finding that though!

4mmonium (approved):
LGTM

PR #373, opened 2023-02-20
Update txAdmin guide on setting up a server

4mmonium:
Updates some of the screenshots for more recent ones and changes server.zip's default link to server.7z including the screenshot.


blattersturm:
This also ought to note that you need a third-party program to open .7z files on Windows, in case users might not know.

4mmonium:
> This also ought to note that you need a third-party program to open .7z files on Windows, in case users might not know.

Ah noticed, I cleared that up, also added download links to the other vanilla setup guide since it was missing such. They were implemented at the bottom, but not referenced anywhere 😃

PR #378, opened 2023-03-11
fix: use of the right variable name (scaleform)

MoskalykA:


4mmonium (commented):
Any explanation as to why this would make a difference? When a method is called, a handle is returned, at least in this specific case.

MoskalykA:
> Any explanation as to why this would make a difference? When a method is called, a handle is returned, at least in this specific case.

I changed this because some functions used the `scaleform` variable without it existing, so I renamed the `ScaleformHandle` variable to correct this

4mmonium (approved):
Thank you for your contribution!

PR #390, opened 2023-06-22
Add build number 2944

4mmonium:
Add build number 2944

### DIFF
--- a/content/docs/server-manual/server-commands.md
+++ b/content/docs/server-manual/server-commands.md
@@ -170,6 +170,7 @@ Every build includes all content and changes from the builds before.
 | 2612   | mpg9ec                               | -                           |
 | 2699   | mpsum2                               | The Criminal Enterprises    |
 | 2802   | mpchristmas3                         | Los Santos Drug Wars        |
+| 2944   | mp2023_01                            | San Andreas Mercenaries     |
 
 **RedM builds**
 


content/docs/server-manual/server-commands.md
```diff
@@ -170,6 +170,10 @@ Every build includes all content and changes from the builds before.
 | 2612   | mpg9ec                               | -                           |
 | 2699   | mpsum2                               | The Criminal Enterprises    |
 | 2802   | mpchristmas3                         | Los Santos Drug Wars        |
+| 2824   | -                                    | -                           |
+| 2845   | -                                    | -                           |
```
blattersturm, line 174: 2824/2845 aren't actually supported by the game, and these docs are meant for end users, not as a general reference.

blattersturm (changes requested):

content/docs/server-manual/server-commands.md
4mmonium: Ah, my bad, I think I should have looked [here](https://github.com/citizenfx/fivem/blob/a01741758e2fac4cbd11d704b1490fadc83f4a7c/code/client/shared/CrossBuildRuntime.h#L8)

I looked at the wrong places [(1)](https://github.com/citizenfx/fivem/blob/a01741758e2fac4cbd11d704b1490fadc83f4a7c/code/components/gta-core-five/include/netPeerAddress.h#L142) [(2)](https://github.com/citizenfx/fivem/blob/a01741758e2fac4cbd11d704b1490fadc83f4a7c/code/components/gta-net-five/src/NetHook.cpp#L1713).

Will correct and commit again.

4mmonium: All done now.
Here's the corrected [commit](https://github.com/citizenfx/fivem-docs/pull/390/files#diff-331e99259d1395f290e695c1c52cfb483254b82a1a1534aadcaabc08ad0d8e5dR173)

content/docs/server-manual/server-commands.md
```diff
@@ -170,6 +170,7 @@ Every build includes all content and changes from the builds before.
 | 2612   | mpg9ec                               | -                           |
 | 2699   | mpsum2                               | The Criminal Enterprises    |
 | 2802   | mpchristmas3                         | Los Santos Drug Wars        |
+| 2944   | mp2023_01, mp2023_01_g9ec            | San Andreas Mercenaries     |
```
Disquse, line 173: I don't think including `mp2023_01_g9ec`is necessary, these are stub DLCs (disabled from loading in FiveM). We never included these in docs for older DLCs before😄

4mmonium: > I don't think including `mp2023_01_g9ec`is necessary, these are stub DLCs (disabled from loading in FiveM). We never included these in docs for older DLCs before😄

Removed stub DLC 😃

PR #394, opened 2023-07-15
Update save_gta_cache

4mmonium:
Explain how to load saved cached data for clients/players, since it was missing from the docs.

### DIFF
--- a/content/docs/client-manual/console-commands.md
+++ b/content/docs/client-manual/console-commands.md
@@ -251,9 +251,27 @@ resource and shows this in a nice overview. Comes in handy when you encounter pe
 Usage: `resmon <true|false>`
 
 ### save_gta_cache
-<!-- TODO: reference a guide on GTA cache and using it in a resource -->
-Saves cache data for a specified resource to the CitizenFX directory in AppData. This is to be used for resources with a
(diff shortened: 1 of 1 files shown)


content/docs/client-manual/console-commands.md
```diff
@@ -251,10 +251,28 @@ resource and shows this in a nice overview. Comes in handy when you encounter pe
...
 
 ### save_gta_cache
-<!-- TODO: reference a guide on GTA cache and using it in a resource -->
-Saves cache data for a specified resource to the CitizenFX directory in AppData. This is to be used for resources with a
+Saves cache data for a specified resource to the CitizenFX directory in AppData (`%AppData%\CitizenFX`). This is to be used for resources with a
 significant amount of collision or map files, to speed up initial loading for players.
 
+In order for players to load this file, it needs to be stored in the server's root folder of the resource you decided to save the map's cache for (not in a subdirectory) and added to your resource manifest.
+
+In your resource manifest, `fxmanifest.lua` (or previously, `__resource.lua`):
```
PsychoShock, line 259: Even though this is right, the documentation should provide the best development practices. Therefore, I suggest removing `__resource.lua`.

4mmonium: You're right, `__resource.lua` is no longer used.

PR #397, opened 2023-09-21
tweak(server-manual/zap): change Patreon wording

Mycroft-Studios:
Current wording makes it seem like OneSync is a paid, Patreon, feature. Which has not been the case since December 2020. The new wording focuses it on the slot count instead, which *is* a Patreon feature


4mmonium (approved):
This seems correct to me. All servers with 'server-side state awareness' are OneSync these days, and it's a feature that is now a given with default server [installs](https://docs.fivem.net/docs/server-manual/setting-up-a-server-vanilla/#servercfg).

PR #403, opened 2023-11-16
tweak: proxy setup guide URL resolution clarification

blattersturm:
This came up in https://forum.cfx.re/t/5185471. I'm not entirely certain about the factual correctness regarding bare hosts trying `http://` first, this should be double-checked with code.

### DIFF
--- a/content/docs/server-manual/proxy-setup.md
+++ b/content/docs/server-manual/proxy-setup.md
@@ -11,9 +11,12 @@ At times, it may be desirable to have a reverse proxy in front of your Cfx.re se
 The client->server connection process is as follows:
 
 1. Client resolves the _connect endpoint_ from the join interaction. This can be any of the following:
-   * The `connectEndPoints` field in the server listing output.
-   * The server address cached by the join URL subsystem.
-   * The raw IP, host or URL entered in a `connect` command.
+   * The `connectEndPoints` field in the server listing output, which can be configured using `sv_listingHostOverride`.
+   * The server address cached by the join URL subsystem, such as from history.
+   * An IP, host or URL entered in a `connect` command. These get processed as follows:
+     * Bare hosts/IPs: `host.example` will resolve to `http://host.example:30120/`.
+     * Hosts with port: `host.example:30121` resolve to `http://host.example:30121/`.
+     * URLs with scheme: `http://host.example/` and `https://host.example/` will be resolved as-is, using the default ports for HTTP and HTTPS (80/443).
 2. Client retrieves general server metadata via a `GET /info.json` on the connect endpoint.
 3. Client attempts an `initConnect` request to the connect endpoint's `POST /client`.
 4. If server accepts client, it'll send a connection token to client.


content/docs/server-manual/proxy-setup.md
```diff
@@ -11,9 +11,12 @@ At times, it may be desirable to have a reverse proxy in front of your Cfx.re se
...
 1. Client resolves the _connect endpoint_ from the join interaction. This can be any of the following:
-   * The `connectEndPoints` field in the server listing output.
-   * The server address cached by the join URL subsystem.
-   * The raw IP, host or URL entered in a `connect` command.
+   * The `connectEndPoints` field in the server listing output, which can be configured using `sv_listingHostOverride`.
+   * The server address cached by the join URL subsystem, such as from history.
+   * An IP, host or URL entered in a `connect` command. These get processed as follows:
+     * Bare hosts/IPs: `host.example` will resolve to `http://host.example:30120/`.
+     * Hosts with port: `host.example:30121` resolve to `http://host.example:30121/`.
+     * URLs with scheme: `http://host.example/` and `https://host.example/` will be resolved as-is, using the default ports for HTTP and HTTPS (80/443).
```
4mmonium, line 19: Looks good to me, I validated this by using `connect` and checking what [`ResolveUrl`](https://github.com/citizenfx/fivem/blob/a44cbfd0b0690a61e8cc65088400b0c4bb3e0c96/code/components/net/src/NetLibrary.cpp#L735) returned. So far this is what I got (gets the [peerAddress](https://github.com/citizenfx/fivem/blob/a44cbfd0b0690a61e8cc65088400b0c4bb3e0c96/code/components/net/src/NetLibrary.cpp#L825)):

| Connect To Argument | Resolves To |
|-------------------------|--------------|
| google.com | http://216.58.204.142:30120/ |
| google.com:30121 |  http://216.58.204.142:30121/ |

Skip right to [here](https://github.com/citizenfx/fivem/blob/a44cbfd0b0690a61e8cc65088400b0c4bb3e0c96/code/components/net/src/NetLibrary.cpp#L762):
| Connect To Argument | Resolves To |
|-------------------------|--------------|
| http://google.com/ | http://google.com/ |
| https://google.com/ | https://google.com/ |
| https://google.com:30121 | https://google.com:30121/ |
| http://google.com:30121 | http://google.com:30121/ |

4mmonium (approved):

content/docs/server-manual/proxy-setup.md
MiguelFGM: Hello @4mmonium @blattersturm, I'm inquiring whether the recent pull request effectively resolved the issue we're discussing. It appears that it only impacts new artifacts without affecting the older ones.

To clarify, in our Reverse Proxy setup as FiveM users, we've already established a subdomain for our server to function as a proxy, using platforms like NGINX or other Content Delivery Networks. In this environment, when 'sv_listingHostOverride' is active, the correct connection should route to port 80/http or 443/https, not the default server port tcp/30120.

A successful connection should look like this:
Connecting to domain.com should redirect to http://domain.com/ or https://domain.com/
It should NOT append port 30120, as in http://domain.com:30120/

Currently, with artifact 6683, the server's connect function isn't operating correctly because it redirects to http://domain.com:30120/ instead of http://domain.com/. However, it works when accessed through the server list. This issue presents a dilemma: maintain the connect command functionality (by disabling the reverse proxy) or face potential DDoS attacks.

Can you confirm if the upcoming updates will address this?

Your assistance is greatly appreciated.

4mmonium: @MiguelFGM Hi, thank you for reporting this. There's no time window as of now, but the team is fully aware of this issue 😊

PR #405, opened 2023-12-05
Update Frameworks

4mmonium:
- Add note letting the user know we're open to new frameworks.
- Add VORP to frameworks and note about ease of deployment when using txAdmin recipes.

### DIFF
--- a/content/docs/server-manual/frameworks.md
+++ b/content/docs/server-manual/frameworks.md
@@ -5,6 +5,7 @@ description: >
   Finding a Framework for your FiveM server.
 ---
 
+# What is a Framework?
 A framework can be utilized to easily make resources for your FiveM server. It provides a barebone structure for your scripts, so you don't have to do everything from scratch. Below are some commonly used frameworks you can use, please note that we are just providing these frameworks for informative purposes, *it's up to you to decide which framework to use*. Additionally, you can search the [releases](https://forum.cfx.re/c/development/releases/7) section, to see if anything new pops up.
 
 #### Running the Frameworks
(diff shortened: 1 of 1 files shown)


content/docs/server-manual/frameworks.md
```diff
@@ -7,7 +7,18 @@ description: >
...
+While the frameworks can be run by following their respective documentation, we might suggest utilizing [txAdmin](https://github.com/tabarra/txAdmin/blob/master/README.md) recipes for enhanced ease in the deployment process. A guide for setting up a server using txAdmin can be found [here][setting-up-a-server-txadmin].
+
+#### Open to new framework suggestions
+If there's a framework you'd like to see included in the list, please submit a request by creating an issue in the documentation [issues][docs-issue-section] section. 
+
 # Frameworks (Lua) (Alphabetical Order)
 - [esx](https://github.com/esx-framework/esx-legacy) ([Docs](https://documentation.esx-framework.org/legacy/installation))
 - [QBCore](https://github.com/qbcore-framework/qb-core) ([Docs](https://docs.qbcore.org/qbcore-documentation/))
+- [VORP](https://github.com/VORPCORE/vorp-core-lua) ([Docs](https://vorpcore.github.io/VORP_Documentation)) (RedM)
 - [vRP](https://github.com/vRP-framework/vRP) ([Docs](https://vrp-framework.github.io/vRP/))
```
PsychoShock, line 20: Probably removing `vRP` from the documentation since this is outdated would be good.

AvarianKnight: That doesn't make much sense as vRP is still widely used.

PR #411, opened 2024-01-09
Fixed file names for vehicle models

glitchdetector:
A few vehicle model file names were not in lowercase, and thus didn't appear in the docs page. 
For consistency, the file names were changed to resolve this.


4mmonium (approved):
Nice catch, totally missed it! 😊

PR #413, opened 2024-01-29
Add playerEnteredScope/playerLeftScope examples

4mmonium:
Adds the implementation examples to the OneSync guide.

### DIFF
--- a/content/docs/scripting-reference/onesync/_index.md
+++ b/content/docs/scripting-reference/onesync/_index.md
@@ -105,7 +105,7 @@ An implementation example can be found down below.
 ## playerEnteredScope
 
 {{% alert title="Performance Warning" color="warning" %}}
-Using these events is frowned upon, these events have scaling performance costs. For every person within the scope of the player on every enter/leave scope this this will be called an additional time, so having 32 players within the scope of a player will lead to this being called 32 times. Whenever possible you should use state bags if you need to trigger scoped events.
+Using these events is frowned upon, these events have scaling performance costs. For every person within the scope of the player on every enter/leave scope this this will be called an additional time, so having 32 players within the scope of a player will lead to this being called 32 times. Whenever possible you should use state bags if you need to trigger scoped events, see {{% native_link "ADD_STATE_BAG_CHANGE_HANDLER" %}}.
 {{% /alert %}}
 
 This event handler is triggered when a player enters another player's scope.
@@ -120,7 +120,7 @@ end)
 ## playerLeftScope
 
 {{% alert title="Performance Warning" color="warning" %}}
-Using these events is frowned upon, these events have scaling performance costs. For every person within the scope of the player on every enter/leave scope this this will be called an additional time, so having 32 players within the scope of a player will lead to this being called 32 times. Whenever possible you should use state bags if you need to trigger scoped events.
+Using these events is frowned upon, these events have scaling performance costs. For every person within the scope of the player on every enter/leave scope this this will be called an additional time, so having 32 players within the scope of a player will lead to this being called 32 times. Whenever possible you should use state bags if you need to trigger scoped events, see {{% native_link "ADD_STATE_BAG_CHANGE_HANDLER" %}}.
 {{% /alert %}}
 
 This event handler is triggered when a player leaves another player's scope.
--- a/content/docs/scripting-reference/onesync/_index.md
+++ b/content/docs/scripting-reference/onesync/_index.md
@@ -104,7 +104,7 @@ An implementation example can be found down below.
 
 ## playerEnteredScope
 
-{{% alert theme="warning" %}}
+{{% alert title="Performance Warning" color="warning" %}}
 Using these events is frowned upon, these events have scaling performance costs. For every person within the scope of the player on every enter/leave scope this this will be called an additional time, so having 32 players within the scope of a player will lead to this being called 32 times. Whenever possible you should use state bags if you need to trigger scoped events.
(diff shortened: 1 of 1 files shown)


AvarianKnight:
These events notoriously are bad to use (bad performance), is it a good idea to publicly document them?

4mmonium:
> These events notoriously are bad to use (bad performance), is it a good idea to publicly document them?

You make a good point, although I think that depends on how the end user writes their implementation. If they write a for loop inside the handler to keep track of users, it's going to be non-performant, if they just use it to assign player related variables i.e. `playerData[playerId].something = true` it may be ok.

Perhaps a word of caution should be added, stating that the method shouldn't be used to perform CPU-intensive operations (such as a for-loop) inside the handler(s) as it will lead to bad performance, for the simple fact that the handler will be triggered every time someone enters/leaves the scope.

Nice catch! 😊

AvarianKnight:
I don't think the issue here is user implementation but the overhead of events, for larger servers using these no matter the circumstance can lead to weird performance issues.

Players should mostly be using state bags if they want to do stuff when a player enters another players scope.

content/docs/scripting-reference/onesync/_index.md
```diff
@@ -97,6 +97,34 @@ Culling is used to avoid sending a lot of synchronization data to the server (su
...
+## playerEnteredScope
+
+This event handler is triggered when a player enters another player's scope.
+
+```lua
+AddEventHandler("playerEnteredScope", function(data)
+    local playerEntered, player = data["player"], data["for"]
+    print(("%s entered %s's scope"):format(playerEntered, player))
+end)
+```
```
AvarianKnight, line 114:
````suggestion

```
````

File: content/docs/scripting-reference/onesync/_index.md [hunk 2]
```diff
@@ -97,6 +97,34 @@ Culling is used to avoid sending a lot of synchronization data to the server (su
...
 When an entity goes out of range, it's no longer controlled by their original owner. This means that any entity that would be out of scope will be culled and migrated/disowned. By default, the culling radius is set to `424 units` around the entity.
 
+# Scopes
+Players may enter/leave other players' scopes, this depends on the culling radius from each other, [server event][server-events] handlers such as `playerEnteredScope` and `playerLeftScope` can be used to track who entered/left someone else's scope.
+
+An implementation example can be found down below.
+
+## playerEnteredScope
+
+This event handler is triggered when a player enters another player's scope.
```
AvarianKnight, line 107:
```suggestion
{{% alert theme="warning" %}}
Using these events is frowned upon, these events have scaling performance costs. For every person within the scope of the player on every enter/leave scope this this will be called an additional time, so having 32 players within the scope of a player will lead to this being called 32 times. Whenever possible you should use state bags if you need to trigger scoped events.
{{% /alert %}}

This event handler is triggered when a player enters another player's scope.
```

content/docs/scripting-reference/onesync/_index.md
```diff
@@ -97,6 +97,34 @@ Culling is used to avoid sending a lot of synchronization data to the server (su
...
+
+```lua
+AddEventHandler("playerEnteredScope", function(data)
+    local playerEntered, player = data["player"], data["for"]
+    print(("%s entered %s's scope"):format(playerEntered, player))
+end)
+```
+## playerLeftScope
+
+This event handler is triggered when a player leaves another player's scope.
```
AvarianKnight, line 117:
```suggestion
{{% alert theme="warning" %}}
Using these events is frowned upon, these events have scaling performance costs. For every person within the scope of the player on every enter/leave scope this this will be called an additional time, so having 32 players within the scope of a player will lead to this being called 32 times. Whenever possible you should use state bags if you need to trigger scoped events.
{{% /alert %}}

This event handler is triggered when a player leaves another player's scope.
```

AvarianKnight (changes requested):

content/docs/scripting-reference/onesync/_index.md
```diff
@@ -97,6 +97,43 @@ Culling is used to avoid sending a lot of synchronization data to the server (su
...
 When an entity goes out of range, it's no longer controlled by their original owner. This means that any entity that would be out of scope will be culled and migrated/disowned. By default, the culling radius is set to `424 units` around the entity.
 
+# Scopes
+Players may enter/leave other players' scopes, this depends on the culling radius from each other, [server event][server-events] handlers such as `playerEnteredScope` and `playerLeftScope` can be used to track who entered/left someone else's scope.
+
+An implementation example can be found down below.
+
+## playerEnteredScope
+
+{{% alert theme="warning" %}}
```
AvarianKnight, line 107:
```suggestion
{{% alert title="Performance Warning" color="warning" %}}
```

content/docs/scripting-reference/onesync/_index.md
```diff
@@ -97,6 +97,43 @@ Culling is used to avoid sending a lot of synchronization data to the server (su
...
+```lua
+AddEventHandler("playerEnteredScope", function(data)
+    local playerEntered, player = data["player"], data["for"]
+    print(("%s entered %s's scope"):format(playerEntered, player))
+end)
+
+```
+## playerLeftScope
+
+{{% alert theme="warning" %}}
```
AvarianKnight, line 122:
```suggestion
{{% alert title="Performance Warning" color="warning" %}}
```

AvarianKnight (changes requested):
I used the wrong formatting here

PR #418, opened 2024-04-15
Update default state for 'sv_filterRequestControl'

tens0rfl0w:
This came up in a recent discussion in the Cfx.re Engineering Group.


4mmonium (approved):
LGTM based on the [current source code](https://github.com/citizenfx/fivem/blob/f35c173a11a60f1a6f4fdaa819b925155bb663d4/code/components/citizen-server-impl/src/state/ServerGameState.cpp#L7223), thank you for your contribution!

PR #421, opened 2024-04-27
feat(section/img): added new section for img natives

PsychoShock:
This will solve https://github.com/citizenfx/natives/pull/1056. People will have the possibility to add new images and reference them with the link.

**Note:** The example image was just to add something since we can't PR a empty folder.

I think this is the easier way to keep track on them.


PsychoShock:
Is this get merged, I'll change the `contribution.md` file so people will know.

PR #422, opened 2024-04-27
feat(doc/script): new doc section for first resource

PsychoShock:
Some credits goes to https://github.com/citizenfx/fivem-docs/pull/201.

This was based on a comment by @technetium-cfx on https://github.com/citizenfx/fivem-docs/pull/201#pullrequestreview-536390489

Nothing "big" but still better than the actual.


PsychoShock:
This is handle in the requested PR - https://github.com/citizenfx/fivem-docs/pull/429

PR #423, opened 2024-04-30
feat(issue/client): add common issues by client

PsychoShock:
This should explain how to handle the Early Exit Trap crash + how to link discord properly.


cubecraft00 (approved):
Added useful, simple and clear solutions

prikolium-cfx:
Is discord issue still happening even after my fix?

PsychoShock:
> Is discord issue still happening even after my fix?

Yes. You can see reports like https://discord.com/channels/192358910387159041/1164137737385934948/1242726213148213273

PsychoShock:
> > Is discord issue still happening even after my fix?
> 
> Yes. You can see reports like https://discord.com/channels/192358910387159041/1164137737385934948/1242726213148213273

Following the question you ask @prikolium-cfx it seems to be when user has a client modification on Discord - I change the section to include that.

content/docs/support/client-issues.md
```diff
@@ -85,6 +85,29 @@ Stuck on a black screen
...
+- Open Discord Desktop App
+- Start FiveM and load main menu
+
+Please avoid using any plugins / client modification when following the steps.
+
+HTTP 412 System Clock
+-----------------------
+Looks like you are not sync on time. Follow those simple steps:
+- Close FiveM and open Windows settings
+- Go to Date & Tim
```
Mathu-lmn, line 105: Small typo here : missed the "e" from time

content/docs/support/client-issues.md
```diff
@@ -85,6 +85,29 @@ Stuck on a black screen
...
+- Sync Clock
+- Start FiveM and load main menu
+
+**Note:** words might change depending on what Windows version you are running (10/11).
```
Mathu-lmn, line 109: "Setting labels" would fit better than just "words"

content/docs/support/client-issues.md
```diff
@@ -85,6 +85,29 @@ Stuck on a black screen
 This is a common issue with certain NVIDIA drivers. Stay patient, it takes a minute to load. This often
 happens to other games too.
 
+Early Exit Trap crash
+-----------------------
+Looks like you might have something messing in the background. Please make sure to remove everything running on background when lauching FiveM. It could be antivirus or any software recording app.
```
Mathu-lmn, line 90: running `in the` background

Mathu-lmn: How can you thumbs down that when it's grammatically correct ? You don't run on background but in the background (as you said in the first sentence ?)

neptunium-cfx: I think this early exit trap mention is unnecessary and it's not a really good explanation on how to 'get rid' of it, a better article exists here https://support.cfx.re/hc/en-us/articles/12505932916508-Game-integrity-check-failed though i get it's not for the same error - we should probably start using it in more cases.

Mathu-lmn: Yeah, tho it's kinda confusing that all client issues are not on the same website :p Why are some in the docs and some in support ? Maybe a good idea would be to merge them to regroup it ?

PsychoShock: > I think this early exit trap mention is unnecessary and it's not a really good explanation on how to 'get rid' of it, a better article exists here https://support.cfx.re/hc/en-us/articles/12505932916508-Game-integrity-check-failed though i get it's not for the same error - we should probably start using it in more cases.

Yeah I knew about this article, sadly it doesnt show up on Google Search that's why I added it in this page. I spoke about this to Tin wich told me to add everything in the same page to be easier.

AvarianKnight:
Waiting on @neptunium-cfx's thoughts here.

PsychoShock:
I'll send him a message to come here.

neptunium-cfx:
I don't think we should document the 'CnL kick' issues. Don't think there's any actual solution that was proven to be working, most of this seem like just 'random' tips. This feature is buggy and will also be corrected in a short future, so that normal users shouldn't experience any issues. For now i'd advise server owners to disable that convar.

content/docs/support/client-issues.md
neptunium-cfx, line 90: Still not super happy about 'It could be antivirus' - might cause some people to disable it, hoping for a quick fix. Do we have any confirmed cases where disabling the AV helps?

neptunium-cfx (changes requested):

content/docs/support/client-issues.md
PsychoShock: Yes, search "disable antivirus" on Discord, you will find many cases where people have disable it and got the "situation" working following this.

content/docs/support/client-issues.md
```diff
@@ -85,6 +85,38 @@ Stuck on a black screen
...
+Can't link FiveM to Discord
+-----------------------
+To be able to connect Discord to FiveM, you need to do the following:
+- Close FiveM and Discord completely
+- Open Discord Desktop App
+- Start FiveM and load main menu
+
+Please avoid using any plugins / client modification when following the steps.
+
+HTTP 412 System Clock
```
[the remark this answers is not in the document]
jellyton255: There are several grammar/spelling errors here that need to be fixed.

PR #424, opened 2024-05-20
docs(natives/five): Add fuel consumption system documentation

Nobelium-cfx:
Documentation for the following set of commits (fuel consumption system implementation):
- https://github.com/citizenfx/fivem/commit/ae65681e39b1bacf5c6fd2d0a98366003fc1d6a6
- https://github.com/citizenfx/fivem/commit/c4d2f2b2d7ce08cc345adc50b5911bd69c041999
- https://github.com/citizenfx/fivem/commit/efd1390c6f9484b9fe57ea8aaddee9a557c6c747
- https://github.com/citizenfx/fivem/commit/4ba5737f9aaf6382de5b103b029c5575e653e162

FYI: links to the new natives may not work because the natives documentation has not propagated to the docs.fivem.net yet.

### DIFF
--- a/content/docs/scripting-manual/using-new-game-features/fuel-consumption.md
+++ b/content/docs/scripting-manual/using-new-game-features/fuel-consumption.md
@@ -124,11 +124,11 @@ function IsPlayerInVehicleAtGasStation(vehicle)
     return false
 end
 
-Citizen.CreateThread(function()
+CreateThread(function()
     -- Main loop.
     while true do
         -- Do not add sleep time since this thread is operating with controls.
-        Citizen.Wait(0)
+        Wait(0)
 
         -- Only run script if fuel consumption is turned on globally.
         if GetFuelConsumptionState() then
--- a/content/docs/scripting-manual/using-new-game-features/fuel-consumption.md
+++ b/content/docs/scripting-manual/using-new-game-features/fuel-consumption.md
@@ -1,0 +1,151 @@
+---
+title: Fuel consumption
+weight: 421
+---
+
+By default in GTA5 and FiveM vehicles do not consume fuel. This feature allows to turn the fuel consumption on and customize it for your needs.
+
+### Turn on/off
+
+To set/check the fuel consumption use [`SET_FUEL_CONSUMPTION_STATE`](https://docs.fivem.net/natives/?_0x81DAD03E)/[`GET_FUEL_CONSUMPTION_STATE`](https://docs.fivem.net/natives/?_0xC66CD90C) natives. Set true to turn it on, false - to turn it off.
+
+### How it works
+
+#### Fuel consumption speed
+
+When turned on, fuel consumption is calculated by the formula:
+
+```
+time_step * revolutions_per_minute * vehicle_fuel_consumption_rate_multiplier * global_fuel_consumption_rate_multiplier
+```
+
+I.e. fuel is consumed faster the more revolutions engine does. This approximates how fuel consumption works in real world.
+
+Fuel is not consumed when the engine is turned off.
+
+By default, 65 litre gas tank car with average fuel consumption can stay **idle for ~16.67 hours** or run with **max RPM for ~2.5 hours**.
+
+#### Customize consumption speed
+
+To customize/check global (across all vehicles) fuel consumption rate (`global_fuel_consumption_rate_multiplier` in the formula above) use [`SET_FUEL_CONSUMPTION_RATE_MULTIPLIER`](https://docs.fivem.net/natives/?_0x845F3E5C)/[`GET_FUEL_CONSUMPTION_RATE_MULTIPLIER`](https://docs.fivem.net/natives/?_0x5550BF9F) natives. By default it is set to 1. If set to negative - 0 will be used instead.
+
+To customize fuel consumption per vehicle (`vehicle_fuel_consumption_rate_multiplier` in the formula above) use [`SET_HANDLING_FLOAT`](https://docs.fivem.net/natives/?_0x90DD01C) (for all vehicles with given class) or [`SET_VEHICLE_HANDLING_FLOAT`](https://docs.fivem.net/natives/?_0x488C86D2) (for a specific vehicle) native with `fieldName` equal to `fPetrolConsumptionRate`. By default it is set to 0.5 for all vehicles.
+
+You can also use [CodeWalker](https://github.com/dexyfex/CodeWalker) tool or similar to edit vehicle `handling.meta` file and set `fPetrolConsumptionRate` value to `HandlingData`. If not set it results to the default value of 0.5.
+
+#### Petrol tank volume and current fuel level
+
+To customize petrol tank volume use [`SET_HANDLING_FLOAT`](https://docs.fivem.net/natives/?_0x90DD01C)/[`SET_VEHICLE_HANDLING_FLOAT`](https://docs.fivem.net/natives/?_0x488C86D2) natives with `fieldName` equal to `fPetrolTankVolume`.
+
+You can also use [CodeWalker](https://github.com/dexyfex/CodeWalker) tool or similar to edit vehicle `handling.meta` file and set `fPetrolTankVolume` value to `HandlingData`.
+
+To update/check fuel level in a vehicle use [`GET_VEHICLE_FUEL_LEVEL`](https://docs.fivem.net/natives/?_0x5F739BB8)/[`SET_VEHICLE_FUEL_LEVEL`](https://docs.fivem.net/natives/?_0xBA970511) natives.
+
+#### Vehicles without fuel consumption
+
+Fuel is not consumed for the following vehicles:
+
+- Vehicles that only have NPCs in them.
+- Bicycles. I.e. vehicles with vehicle type equal to 12.
+- Vehicles with infinite fuel. I.e. vehicles with petrol tank volume equal to 0. By default it's only bicycles.
+
+To check if vehicle will consume fuel when player is inside (i.e. petrol tank volume above 0 and not a bicycle) use [`DOES_VEHICLE_USE_FUEL`](https://docs.fivem.net/natives/?_0xEF30A696) native.
+
+### Gas stations
+
+When turning on fuel consumption you need to think about mechanisms to allow players to refuel a vehicle.
+
+We do not provide out of the box functionality for gas stations. But you can implement it on your own using provided natives.
+
+#### Example gas stations implementation
+
+```lua
+-- List of all gas stations in the world.
+-- Update this list to include all gas station locations.
+-- You can create a config file and read it from there, or just hard code as in this example.
+GasStations = {
+    {
+        coords = vector3(64.55, 20.4, 68.9),  
+        radius = 8
+    }
+}
+
+-- Distance from point is the simplest way to determine if vehicle is at a gas station.
+-- You can use polygons if you need more precision.
+function IsPointInGasStation(point, gasStation)
+    return #(point - gasStation.coords) <= gasStation.radius
+end
+
+-- Triggered when player enters gas station.
+-- You can customize communication with a player the way you want.
+function ProcessGasStationEnter(vehicle)
+    -- Get current vehicle fuel level.
+    local vehicleFuelLevel = GetVehicleFuelLevel(vehicle)
+    -- Get vehicle petrol tank volume.
+    local vehicleGasTankVolume = GetVehicleHandlingFloat(vehicle, "CHandlingData", "fPetrolTankVolume")
+    -- Display basic text message.
+    local gasStationMessage = string.format("Welcome to gas station. You have %.3f out of %.3f liters of fuel left. Press G to fill tank.", vehicleFuelLevel, vehicleGasTankVolume)
+    AddTextEntry("CH_ALERT", gasStationMessage)
+    BeginTextCommandDisplayHelp("CH_ALERT")
+    EndTextCommandDisplayHelp(0, false, false, 200)
+end
+
+-- Fill up player's vehicle gas tank.
+-- You can customize communication with a player the way you want.
+function ProcessGasStationFuelPurchase(vehicle)
+    -- Get vehicle petrol tank volume.
+    local vehicleGasTankVolume = GetVehicleHandlingFloat(vehicle, "CHandlingData", "fPetrolTankVolume")
+    -- Fill up the tank.
+    SetVehicleFuelLevel(vehicle, vehicleGasTankVolume)
+end
+
+function IsPlayerInVehicleAtGasStation(vehicle)
+    local playerPed = PlayerPedId()
+    if not playerPed or playerPed == -1 then
+        return false
+    end
+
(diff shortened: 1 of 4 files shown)


content/docs/scripting-manual/using-new-game-features/fuel-consumption.md
```diff
@@ -0,0 +1,151 @@
...
+        end
+    end
+
+    return false
+end
+
+Citizen.CreateThread(function()
+    -- Main loop.
+    while true do
+        Citizen.Wait(50)
```
Disquse, line 133: Given that we are displaying a help text in this thread and operating with controls, this must be `Citizen.Wait(0)`

content/docs/scripting-manual/using-new-game-features/fuel-consumption.md
```diff
@@ -0,0 +1,151 @@
...
+
+function IsPlayerInVehicleAtGasStation(vehicle)
+    local playerPed = PlayerPedId()
+    if not playerPed or playerPed == -1 then
+        return false
+    end
+
+    -- If a player is not in a vehicle.
+    local vehicle = GetVehiclePedIsIn(playerPed, false)
+    if not vehicle or not DoesEntityExist(vehicle) then
```
Disquse, line 110: An invalid vehicle will return script guid `0`. This must be `if vehicle == 0 or not DoesEntityExist(vehicle) then` as `0` as means `true` in Lua.

content/docs/scripting-manual/using-new-game-features/fuel-consumption.md
```diff
@@ -0,0 +1,151 @@
...
+function ProcessGasStationFuelPurchase(vehicle)
+    -- Get vehicle petrol tank volume.
+    local vehicleGasTankVolume = GetVehicleHandlingFloat(vehicle, "CHandlingData", "fPetrolTankVolume")
+    -- Fill up the tank.
+    SetVehicleFuelLevel(vehicle, vehicleGasTankVolume)
+end
+
+function IsPlayerInVehicleAtGasStation(vehicle)
+    local playerPed = PlayerPedId()
+    if not playerPed or playerPed == -1 then
```
Disquse, line 104: An invalid player ped will have `0` guid, can't ever be `-1`. This check should look something like that:
`if playerPed == 0 then`.

Disquse (changes requested):
Overall looks good, but there are slight changes I would like to see in the attached script (just to not give a bad example).

content/docs/scripting-manual/using-new-game-features/fuel-consumption.md
Nobelium-cfx: Text is displayed with some duration - so I it works file. But good catch about controls! Fixed.

content/docs/scripting-manual/using-new-game-features/fuel-consumption.md
Nobelium-cfx: Oh, wow. Thanks! Fixed.

content/docs/scripting-manual/using-new-game-features/fuel-consumption.md
Nobelium-cfx: Done.

PsychoShock: Instead of checking twice the value, why not

`if (playerPed or 0) <= 0 then`

PsychoShock (changes requested):
This PR adheres to American English (AmE) conventions with the following minor exceptions:

Spelling:

- "litre" should be changed to "liter" for American English.
- "litres" should be changed to "liters" for American English.

Punctuation:

- Ensure consistent use of the em dash (—) or en dash (–) with spaces for readability in lists or instructions. In American English, dashes are often used without spaces.

Following method signature, is `Citizen.` now used? (Even if doesnt change anything) a method signature should be set to standardize everything.

Also, the example should disable key `58` instead of checking if pressed.

This is all my personnal opinion.

content/docs/scripting-manual/using-new-game-features/fuel-consumption.md
Nobelium-cfx: I think `if (playerPed or 0) <= 0 then` is harder to understand and I don't see any advantages in comparison to the explicit check of two return codes.

content/docs/scripting-manual/using-new-game-features/fuel-consumption.md
```diff
@@ -0,0 +1,152 @@
...
+function ProcessGasStationFuelPurchase(vehicle)
+    -- Get vehicle petrol tank volume.
+    local vehicleGasTankVolume = GetVehicleHandlingFloat(vehicle, "CHandlingData", "fPetrolTankVolume")
+    -- Fill up the tank.
+    SetVehicleFuelLevel(vehicle, vehicleGasTankVolume)
+end
+
+function IsPlayerInVehicleAtGasStation(vehicle)
+    local playerPed = PlayerPedId()
+    if playerPed == 0 or playerPed == -1 then
```
AvarianKnight, line 104: Is there even a case when PlayerPedId can return an invalid ped?

Nobelium-cfx: Hm, idk, maybe not. @Disquse WDYT?

slashkeyvalue: Yes, i think you should just keep it simple and use `DoesEntityExist` instead of any of these checks.

Nobelium-cfx: Yea, I think you are right. None of our other examples do those checks either. Not sure why I added it at the first place. Removed.

Nobelium-cfx:
@PsychoShock , thanks for suggestions! 

> This PR adheres to American English (AmE) conventions with the following minor exceptions:
> 
> Spelling:
> 
> * "litre" should be changed to "liter" for American English.
> * "litres" should be changed to "liters" for American English.

Updated the `litre` one to `liter`. It was also inconsistent within this doc (plural one was `liters`).

> Punctuation:
> 
> * Ensure consistent use of the em dash (—) or en dash (–) with spaces for readability in lists or instructions. In American English, dashes are often used without spaces.

Can you elaborate? I`m not sure what do you mean. Sorry I'm really not an expert at this :). 

> Following method signature, is `Citizen.` now used? (Even if doesnt change anything) a method signature should be set to standardize everything.

Good point. I was following an example of the default script that spawns player. But its actually inconsistent there as well. Let's not add the prefix I guess. Removed.

> Also, the example should disable key `58` instead of checking if pressed.

Do you suggest to set action on a button and then disable the button if condition is not met? I'm not sure if it's any better. I may be missing something tho.

content/docs/scripting-manual/using-new-game-features/fuel-consumption.md
```diff
@@ -0,0 +1,149 @@
...
+
+Citizen.CreateThread(function()
+    -- Main loop.
+    while true do
+        -- Do not add sleep time since this thread is operating with controls.
+        Citizen.Wait(0)
+
+        -- Only run script if fuel consumption is turned on globally.
+        if GetFuelConsumptionState() then
+            if IsPlayerInVehicleAtGasStation() then
```
AvarianKnight, line 135: Maybe this function should be refactored to use `GetVehiclePedIsIn` and pass it to `IsPlayerInVehicleAtGasStation` so that there aren't repetitive native calls

```lua
-- checks if the specified vehicle is close to a gas station
-- caller must ensure that the vehicle exists
function IsVehicleAtGasStation(vehicle)
    -- If the vehicle doesn't use fuel (i.e. a bicycle or with infinite fuel).
    if not DoesVehicleUseFuel(vehicle) then
        return false
    end

    -- Get coordinates of the vehicle and check that it's close enough to any of the gas stations.
    local vehicleCoords = GetEntityCoords(vehicle)
    for i = 1, #gasStations do
    	 if IsPointInGasStation(vehicleCoords, gasStations[i]) then
            return true
        end
    end
    return false
end
```

And refactor `CreateThread` to

```lua
CreateThread(function()
    -- Main loop.
    while true do
        -- Do not add sleep time since this thread is operating with controls.
        Wait(0)

		local playerPed = PlayerPedId()
        -- Only run script if fuel consumption is turned on globally.
        if GetFuelConsumptionState() then
        	local vehicle = GetVehiclePedIsIn(PlayerPedId(), false)
        	
     		-- GetVehiclePedIsIn shouldn't return an invalid vehicle, only 0 if the vehicle doesn't exist.
        	if vehicle ~= 0 and IsVehicleAtGasStation(vehicle) then
                ProcessGasStationEnter(vehicle)

                -- If G is pressed on a keyboard.
                if IsControlPressed(0, 58) then
                    ProcessGasStationFuelPurchase(vehicle)
                end
        	end
        end
    end
end)
```

thelindat (commented):
Maybe not important or worthwhile, but perhaps for the sake of following "good practices" the examples should use local variables and functions?

PsychoShock:
> > Punctuation:
> > 
> > * Ensure consistent use of the em dash (—) or en dash (–) with spaces for readability in lists or instructions. In American English, dashes are often used without spaces.
> 
> Can you elaborate? I`m not sure what do you mean. Sorry I'm really not an expert at this :).
> 
Yes, talking about
"Set true to turn it on, false - to turn it off." should be "Set true to turn it on, false—to turn it off." 
In American English, it's more common to use an em dash (—) without spaces around it for such cases

> > Following method signature, is `Citizen.` now used? (Even if doesnt change anything) a method signature should be set to standardize everything.
> 
> Good point. I was following an example of the default script that spawns player. But its actually inconsistent there as well. Let's not add the prefix I guess. Removed.

Yeah just saw that, I'll PR the docs for docs signature not to use prefixs.

> > Also, the example should disable key `58` instead of checking if pressed.
> 
> Do you suggest to set action on a button and then disable the button if condition is not met? I'm not sure if it's any better. I may be missing something tho.

No, just disable the key direclty in the thread and check if control disable is pressed. Adding buttons in a table is good when multiples buttons are used.

AvarianKnight:
> "Set true to turn it on, false - to turn it off." should be "Set true to turn it on, false—to turn it off."
> In American English, it's more common to use an em dash (—) without spaces around it for such cases

This seems pretty useless, imo, these are all things a lay person wont notice, this isn't getting written on for some academic thesis

content/docs/scripting-manual/using-new-game-features/fuel-consumption.md
```diff
@@ -0,0 +1,149 @@
...
+end
+
+Citizen.CreateThread(function()
+    -- Main loop.
+    while true do
+        -- Do not add sleep time since this thread is operating with controls.
+        Citizen.Wait(0)
+
+        -- Only run script if fuel consumption is turned on globally.
+        if GetFuelConsumptionState() then
```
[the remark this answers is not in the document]
4mmonium: That's the native [`GET_FUEL_CONSUMPTION_STATE`](https://github.com/citizenfx/fivem/blob/64c154cd4cb63dd2ec8407f60f9f34919fbe0726/ext/native-decls/GetFuelConsumptionState.md?plain=1#L8) which isn't visible in the native reference list yet. 🙂

content/docs/scripting-manual/using-new-game-features/fuel-consumption.md
Disquse: From what I remember, GTA5 doesn't support players without peds and will crash as soon as you remove your ped, unlike RDR3 (at least from my last test). So, this scenario is extremely unlikely to happen. Given that all further native calls will return 0 if you pass an invalid ped script GUID, it doesn't make much sense to check for player ped existence.

content/docs/scripting-manual/using-new-game-features/fuel-consumption.md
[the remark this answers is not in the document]
Nobelium-cfx: Yes. You can see PR that added this native attached in the description. There is an issue with the natives documentation update - it is being worked on IIUC.

Nobelium-cfx:
@PsychoShock 

> "Set true to turn it on, false - to turn it off." should be "Set true to turn it on, false—to turn it off."
> In American English, it's more common to use an em dash (—) without spaces around it for such cases

I see. I actually don't see us using em dash like this. We seem to always use " - ". So lets keep it as it is to keep it consistent.

> No, just disable the key direclty in the thread and check if control disable is pressed. Adding buttons in a table is good when multiples buttons are used.

Sorry, I still don't get it. Can you give an example?

content/docs/scripting-manual/using-new-game-features/fuel-consumption.md
Nobelium-cfx: Sounds good. Done.

content/docs/scripting-manual/using-new-game-features/fuel-consumption.md
```diff
@@ -132,15 +125,19 @@ CreateThread(function()
 
         -- Only run script if fuel consumption is turned on globally.
         if GetFuelConsumptionState() then
-            if IsPlayerInVehicleAtGasStation() then
 
-                local vehicle = GetVehiclePedIsIn(PlayerPedId(), false)
+            -- Check if player is in a vehicle.
+            local vehicle = GetVehiclePedIsIn(PlayerPedId(), false)
+            if DoesEntityExist(vehicle) then
```
AvarianKnight, line 131: You can just directly compare that the vehicle isn't 0 here as the vehicle will either be valid or 0

Nobelium-cfx:
> Maybe not important or worthwhile, but perhaps for the sake of following "good practices" the examples should use local variables and functions?

@thelindat I think I do mark all local variables as local already. The `GasStations` one is intended to be global. Ideally it would not be hard coded at all, but I think it would make the example overly complicated.

As for the functions - do you suggest declaring all functions as local in the main thread function? I'm not sure if it's a good idea. Or did I misunderstand you?

content/docs/scripting-manual/using-new-game-features/fuel-consumption.md
Nobelium-cfx: I think using `DoesEntityExist` is more expressive. Also example in https://docs.fivem.net/natives/?_0x9A9112A0FE9A4713 uses it as well

thelindat:
> do you suggest declaring all functions as local in the main thread function? I'm not sure if it's a good idea.

It's just standard convention to avoid globals unless necessary; though it's generally confusing for beginners who don't really understand scope.

Like I said, maybe not important - just a nitpick since everyone else was doing it. 🙃

Nobelium-cfx:
> It's just standard convention to avoid globals unless necessary; though it's generally confusing for beginners who don't really understand scope.
> 
> Like I said, maybe not important - just a nitpick since everyone else was doing it. 🙃

No problem, suggestions are always welcome :). 

But I think global functions are actually more intuitive and easier to understand. I will keep this one as it is.

Disquse (approved):

PR #431, opened 2024-06-05
feat(wait/doc): complete page

PsychoShock:
This will mostly show:
- What is Citizen.Wait
- How does it work
- Example
- Where to use it
- Tips


TheIndra55:
In my opinion this has too much filler, while leaving out important details like that it yields "thread"/coroutine so the game can render the next frame. You should probably just extend the current page and include two examples; one example showing a 2 seconds wait, and another one explaining how/why we need the `Wait(0)` you see often in loops.

PsychoShock:
Credits to @AvarianKnight for the "Things to keep in mind" part. Thanks.

content/docs/scripting-reference/runtimes/lua/functions/Citizen.Wait.md
```diff
@@ -2,19 +2,91 @@
...
+
+If a player is running at 60 fps they will have a tick happen once every 16.6ms.
+If a player is running at 180 fps they will have a tick happen once every 5.5ms.
+--
+
+`Wait(2000)` will pause the current `[coroutine](https://en.wikipedia.org/wiki/Coroutine)` for approximately 2000 ms (2 seconds).
+
+## Best Practices
+
+1. **Avoid short waits**: Be cautious about using short waits in threads such as `0ms` for things that can take a higher time.
```
AvarianKnight, line 86: The wording here might lead people to believe that they should use `Wait(5)` and other "weird" wait times trying to "optimize" their scripts.

This should probably more so say "use different coroutines to cache values that don't need to be called every frame, and don't call expensive native calls per-frame, like [START_EXPENSIVE_SYNCHRONOUS_SHAPE_TEST_LOS_PROBE](https://docs.fivem.net/natives/?_0x377906D8A31E5586), if something needs to be run per-frame run it with `Wait(0)`."

content/docs/scripting-reference/runtimes/lua/functions/Citizen.Wait.md
```diff
@@ -2,19 +2,90 @@
 title: Citizen.Wait
 ---
 
-This will "pause" the execution of the current thread for miliseconds amount of time.
+## What is Citizen.Wait?
 
-Syntax
-------
+`Citizen.Wait` is a function in FiveM that allows you to pause the execution of the current script thread for a specified amount of time. This can be useful for creating delays, scheduling tasks, or controlling the flow of your script.
```
AvarianKnight, line 7:
```suggestion
`Citizen.Wait` is a function in FiveM that allows you to pause the execution of the current thread for a specified amount of time. This can be useful for creating delays, scheduling tasks, or controlling the flow of your script.
```

content/docs/scripting-reference/runtimes/lua/functions/Citizen.Wait.md
```diff
@@ -2,19 +2,90 @@
...
+`Wait(2000)` will pause the current `[coroutine](https://en.wikipedia.org/wiki/Coroutine)` for approximately 2000 ms (2 seconds).
+
+## Best Practices
+
+1. **Avoid short waits**: use different coroutines to cache values that don't need to be called every frame, and don't call expensive native calls per-frame, like [START_EXPENSIVE_SYNCHRONOUS_SHAPE_TEST_LOS_PROBE](https://docs.fivem.net/natives/?_0x377906D8A31E5586) if it can be avoided, if something needs to be run per-frame run it with `Wait(0)`.
+2. **Combine with Conditional Logic**: You can combine `Citizen.Wait` with conditional logic to create more sophisticated behaviors. Please refer to the sleep method example.
 
-Examples
---------
+## Conclusion
```
AvarianKnight, line 89:
```suggestion

```

content/docs/scripting-reference/runtimes/lua/functions/Citizen.Wait.md
```diff
@@ -2,19 +2,90 @@
...
 
-TODO
+`Citizen.Wait` is a simple function in FiveM scripting that allows you to control the timing and flow of your scripts. By pausing execution for a specified duration, you can create delays, schedule tasks, and manage the sequence of actions in your code. Whether you're creating simple delays or orchestrating complex task sequences, `Citizen.Wait` is an essential tool for effective script management.
```
AvarianKnight, line 91:
```suggestion

```

content/docs/scripting-reference/runtimes/lua/functions/Citizen.Wait.md
```diff
@@ -2,19 +2,90 @@
...
+```lua
+Citizen.CreateThread(function()
+    while true do
+        print("This message will print every 5 seconds.")
+        Citizen.Wait(5000) -- Wait for 5000 milliseconds (5 seconds)
+    end
+end)
+```
+
+In this example, the script will continuously print a message every 5 seconds. Note that not waiting at least `0ms` in a thread will crash the client game.
```
AvarianKnight, line 46:
```suggestion
In this example, the script will continuously print a message every 5 seconds.

{{% alert color="warning" %}}
Doing a `while true do` loop without a `Wait` will crash the clients game, please make sure your logic doesn't accidentally prevent `Wait` from being called. 
{{% /alert %}}
```

AvarianKnight (changes requested):

PR #435, opened 2024-06-06
feat(developpers/docs): setup guidelines

PsychoShock:


content/docs/contributing/how-you-can-help.md
```diff
@@ -2,34 +2,33 @@
 title: How you can help
 ---
 
-There are many ways you can help FiveM in its development.
+# How You Can Help
```
4mmonium, line 5: Adding this will make the title show twice, since by default articles use the provided title given up above in this case the one at line 2.

PsychoShock: Ah yes, good catch. Fix.

content/docs/contributing/how-you-can-help.md
```diff
@@ -2,34 +2,32 @@
...
+## Improving Documentation
 
-Documentation is an ever-ongoing piece of work. If you see any room for improvement on these docs, you can do so by
-submitting a pull request on our [docs repository][docs-rep], or clicking the "Improve this page" link at the bottom of
-a page.
+Documentation is a continual work in progress, and we always welcome improvements. If you spot any areas for improvement, you can submit a pull request on our [docs repository][docs-rep], or simply click the "Improve this page" link at the bottom of any documentation page. You can also create a Github issue to suggest new pages. Any contribution is welcome!
 
-Patreon
--------
+### Method signatures
```
PsychoShock, line 19: Just also adding a method signature since it's important to keep a consistency accross all docs.

content/docs/contributing/how-you-can-help.md
```diff
@@ -2,34 +2,32 @@
...
+
+Thank you for your support and contributions!
 
 [developer-docs]: /docs/developers
 [discord]: https://discord.gg/fivem
-[server-data]: https://github.com/citizenfx/cfx-server-data
 [docs-rep]: https://github.com/citizenfx/fivem-docs
 [patreon]: https://patreon.com/fivem
+
+---
```
AvarianKnight, line 33:
```suggestion

```
I don't think this is needed for anything.

AvarianKnight (changes requested):
Only one thing that I notice here.

PR #439, opened 2024-06-06
refactor(perms/tx): update list

PsychoShock:
All credits to https://github.com/tabarra/txAdmin/blob/master/docs/permissions.md


4mmonium (approved):
LGTM, thank you!

Issue #417, opened 2024-03-07
Better explanations to build fivem docs


ahcenezdh:
Currently, the fivem-docs uses a [fork](https://github.com/blattersturm/hugo.git) of Hugo. On GitHub, there are no binaries provided, nor instructions on how to build them.

It would be beneficial to provide a binary for everyone or explain the build process.

Here is a .bat script I created for this purpose. Before running it, you need to install Go and GCC (note that Cygwin will not work; MinGW 64-bit is recommended).

This script will build the Hugo binary (hugo.exe)

```bat
@echo off
REM Git clone here
git clone https://github.com/blattersturm/hugo.git

REM CD Folder
cd hugo

REM SWITCH BRANCH
git switch processed-content-variable # switch on the good branch

REM SET GO
go env -w CGO_ENABLED=1

REM Install Extended
go install --tags extended # if you don't have GCC won't work

REM Build Hugo
go build --tags extended # if you don't have GCC won't work
```

PR #443, opened 2024-06-07
feat(page/doc): custom 404 page

PsychoShock:
Should be handle following https://gohugo.io/templates/404/

Will also handle Github issue if someone found a invalid link

Issue is people that goes into 404 page can't "return home".

image

### DIFF
--- a/layouts/404.html
+++ b/layouts/404.html
@@ -1,0 +1,9 @@
+{{ define "main" }}
+  <h1>404 | Page not found</h1>
+  <p>If you got redirected from a doc page, please report this issue <a href="https://github.com/citizenfx/fivem-docs/issues/new">here</a>.</p>
+  <p>
+    <a href="https://docs.fivem.net/docs/">
+      Return to the home page
+    </a>
+  </p>
+{{ end }}


AvarianKnight:
This sadly does not work.

PsychoShock:
I wasn't sure. I spoke with some membrs years ago about this. It was a custom template that was suppose to be working. Sadly building docs are broken so no way to find out.

AvarianKnight:
Building the docs works fine if you're using the custom fork.

You just have to make sure you build with `go build --tags extended`

ahcenezdh:
> I wasn't sure. I spoke with some membrs years ago about this. It was a custom template that was suppose to be working. Sadly building docs are broken so no way to find out.

It's not broken.

See #417

AvarianKnight:
Did a bit more research, this is not broken the version of hugo we're on doesn't support showing the 404.html on localhost

Reference: https://discourse.gohugo.io/t/why-do-custom-404-pages-not-work-on-hugo-localhost/39227

AvarianKnight:
This doesn't fix the 404 page for the UI though, sadly.

PR #444, opened 2024-06-07
feat(ipl/docs): create game reference

PsychoShock:
Since it's prefer to have everything in docs -- also easier to update them if needed.

All credits goes to @4mmonium for this list.

Please note that this also create a "natives" section to easily add more files there and avoir organization confusion.

This is related to


content/docs/game-references/natives/ipl-list.md
```diff
@@ -0,0 +1,1198 @@
+---
+title: HUD colors
+toc_hide: true
```
4mmonium, line 3: Hi, HUD colors here, also `toc_hide`? 🤔

4mmonium (changes requested):
Please check my comments, I also think this should be shown in the navigation pane, thanks!

AvarianKnight:
Since these are presumably supposed to be usable in game should these not be in an easier to copy format?

4mmonium:
> Since these are presumably supposed to be usable in game should these not be in an easier to copy format? Having X Y Z in different columns vs just having `vector(x, y ,z)` makes this a more annoying process than it needs to be

Indeed, I should probably reformat these to at least be `X, Y, Z` in a single column.

PsychoShock:
Yeah I'll just wait for you to reformat them then. I think it will be much easier.

colistro123:
> Yeah I'll just wait for you to reformat them then. I think it will be much easier.

This has been updated, also broken down into categories and sub-categories. Matching types can be found below, names are also ordered alphabetically. All available sub-categories can be found below.

https://gist.github.com/4mmonium/2fa9aa87f3b02ee6198766c83413c277

| Matching Type | Category Sub-Description |
|---------------|---------------------------|
| burnt         | Burnt                     |
| unburnt       | Unburnt                   |
| open          | Open                      |
| closed        | Closed                    |
| normal        | Normal                    |
| explosion     | Explosion                 |
| crash         | Crash                     |
| fire          | Fire                      |
| sunk          | Sunk                      |
| interior      | Interior                  |
| destroyed     | Destroyed                 |
| fixed         | Fixed                     |
| clean         | Clean                     |
| trash         | Trash                     |
| broken        | Broken                    |

PR #451, opened 2024-06-13
feat(section/resources): example resources docs

PsychoShock:
This is simple pages following https://github.com/citizenfx/example-resources comments to add useful informations instead of adding lot of comments on the resources it-self. Comments will be able to reference directly on docs page.

### DIFF
--- a/content/docs/resources/example-resources/_index.md
+++ b/content/docs/resources/example-resources/_index.md
@@ -1,0 +1,29 @@
+---
+title: Example resources
+---
+
+## About
+These resources adds a variety of gamemodes. They also cover the following topics:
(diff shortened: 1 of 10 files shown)


AvarianKnight:
Keeping these in the example resources would probably make more sense than splitting the documentation.

PsychoShock:
"FiveM" always supported the resources [made by them](https://docs.fivem.net/docs/resources/) in [docs](https://docs.fivem.net/docs/resources/) so having them to handle the `example-resource` (that no one knows it exist) make sense instead of closing it. If it goes on that way, then scripting resources section should get removed on docs too.

PR #452, opened 2024-06-14
tweak(player/js): update player ped id

PsychoShock:
Replacing old native name.


4mmonium (approved):
LGTM since use of `GetPlayerPed(-1)` is discouraged.

PR #457, opened 2024-07-09
feat(doc): add security guideline

PsychoShock:
This will be a good "start" for people building security in their server.

### DIFF
--- a/content/docs/developers/server-security.md
+++ b/content/docs/developers/server-security.md
@@ -1,0 +1,209 @@
+---
+title: Make your server more secure
+draft: true
+---
+
+Our anti-cheat cannot protect some of the exploit your server might have. In this simple guide, we'll try to show you how to get started with some tips to prevent in Lua. 
+
+## Events
+Some cheats authorize the client to trigger events allowing them to have access to a wide possibilities, let's see how you can protect them:
+
+### Proper Use of Event Handlers in Lua
+When working with events in Lua, it's crucial to register them correctly based on whether they are called by the client or the server. A common mistake is registering server events that are not supposed to be called by the client, or vice-versa, which can lead to security vulnerabilities. Here’s a concise guide to using `AddEventHandler` and `RegisterNetEvent` appropriately:
+
+#### `AddEventHandler`
+
+Use `AddEventHandler` when the event is intended to be triggered within the same context, either client-client or server-server. This ensures that the events are not networked and cannot be called by the opposing side.
+
+```lua
+AddEventHandler("eventName", function(eventParam1, eventParam2)
+    -- Code here will be executed once the event is triggered within the same context.
+end)
+```
+
+#### `RegisterNetEvent`
+
+Use `RegisterNetEvent` when the event needs to be triggered across different contexts, such as from client to server or server to client.
+
+**NOTE**: This doesn't block execution from the same context, under the hood `RegisterNetEvent` is a wrapper that is just the following
+
+```lua
+RegisterNetEvent("eventName")
+AddEventHandler("eventName", function()
+
+end)
+```
+
+```lua
+RegisterNetEvent("eventName", function(eventParam1, eventParam2)
+    -- Code here will be executed once the event is triggered across different contexts.
+end)
+```
+
+You can learn more [on this guide](docs/scripting-manual/working-with-events/listening-for-events/).
+
+### Add Checks
+Even if you build a strong anti-cheat, adding checks on **server** events makes them more secure. This is a good practice, although it doesn't prevent everything. Below, we will share some good tips.
+
+When using networked events, make sure to add some checks like:
+- Player money
+- Player [state bags](/docs/scripting-manual/networking/state-bags/)
+- Player inventory items
+- Player [position](https://docs.fivem.net/natives/?_0x1647F1CB)
+- Player experience and level
+- Player [permissions](https://forum.cfx.re/t/basic-aces-principals-overview-guide/90917) and roles
+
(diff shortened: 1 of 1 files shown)


content/docs/developers/server-security.md
```diff
@@ -0,0 +1,239 @@
...
+
+### Results on Player
+Having those convars activate will likely get the player kick with the following reason:
+```
+Connection to CNL timed out.
+```
+
+**Note:** some error *might occur* but it's unlikely. 
+
+User getting kick doesn't mean is automatically Global Ban. However, it provides an indication of player reliability, which can be useful for assessing trustworthiness.
```
Mathu-lmn, line 229: User getting kick**ed** doesn't mean **he** is automatically globally banned

content/docs/developers/server-security.md
```diff
@@ -0,0 +1,239 @@
...
+
+We also have some other options:
+
+```
+sv_disableClientReplays
+```
+Enabling this will aim to reduce chances of cheating options. Please note that this will disable Rockstar Editor.
+
+## Important to know
+The codes provided are not suppose to be working on a copy/paste method. This is just some tips to prevent some actions that might happen in the server. This require some knowledge, you are always free to join our [discord](discord.gg/fivem) to get additionnal help.
```
Mathu-lmn, line 239: suppose**d**

content/docs/developers/server-security.md
```diff
@@ -0,0 +1,239 @@
+---
+title: Make your server more secure
+draft: true
+---
+
+Our anti-cheat cannot protect some of the exploit your server might have. In this simple guide, we'll try to show you how to get started with some tips to prevent in Lua. 
+
+## Events
+Some cheats authorize the client to trigger events allowing them to have access to a wide possibilities, let's see how you can protect them:
```
Mathu-lmn, line 9: Could be interesting mentioning event honeypots too

content/docs/developers/server-security.md
```diff
@@ -0,0 +1,239 @@
...
+
+When using networked events, make sure to add some checks like:
+- Player money
+- Player [state bags](/docs/scripting-manual/networking/state-bags/)
+- Player inventory items
+- Player [position](https://docs.fivem.net/natives/?_0x1647F1CB)
+- Player experience and level
+- Player [permissions](https://forum.cfx.re/t/basic-aces-principals-overview-guide/90917) and roles
+
+Make sure to retrieve all values using server-side methods, not allowing players to change the values. This ensures the integrity and security of your game environment. Please note that client checks can also be a good pratice but can be easily be ovverided.
```
Mathu-lmn, line 47: overridden instead of ovverided

content/docs/developers/server-security.md
```diff
@@ -0,0 +1,239 @@
...
+### Potential Consequences of SQL Injection
+
+- **Unauthorized Access**: Attackers can bypass authentication mechanisms and gain unauthorized access to user data.
+- **Data Manipulation**: Attackers can modify, delete, or insert data, leading to data corruption or loss.
+- **System Compromise**: If the database contains sensitive information or controls critical parts of the application, a successful SQL injection can lead to a complete system compromise.
+
+To mitigate the risks of SQL injection and ensure the security, it is essential to use parameterized queries [(as shown in Example 1)](#Example1). This practice effectively prevents the injection of malicious SQL code and protects the database from unauthorized access and manipulation.
+
+## Cheaters
+Even thought we are trying to keep our community safe, some users might have not been detected yet. This is some tools to prevents some popular actions:
```
Mathu-lmn, line 92: though not thought :p
`There are some tools to prevent` instead of `This is some tools to prevents`

Mathu-lmn (changes requested):
Good start overall :)

AvarianKnight:
I don't think that resource specific exploitation/mitigations should be included in the fivem docs.

Some of this would probably be better off living in a modding tutorial and be linked to from this doc.

@PsychoShock thoughts?

PsychoShock:
> I don't think that resource specific exploitation/mitigations should be included in the fivem docs.
> 
> Some of this would probably be better off living in a modding tutorial and be linked to from this doc.
> 
> @PsychoShock thoughts?

I had the same thought—although the resource is extremely popular, this document is more about *demonstrating* how to prevent general situations rather than focusing on this specific resource. While I agree that a modding tutorial linked to the main docs could work, it would lead users to multiple documents (unless you have a better option). Perhaps I could remove the specific examples and keep the guidance more general?

Do you have anything in mind?

AvarianKnight:
> While I agree that a modding tutorial linked to the main docs could work, it would lead users to multiple documents (unless you have a better option). Perhaps I could remove the specific examples and keep the guidance more general?

I think this is a good idea, just remove the resource specific example

PsychoShock:
> > While I agree that a modding tutorial linked to the main docs could work, it would lead users to multiple documents (unless you have a better option). Perhaps I could remove the specific examples and keep the guidance more general?
> 
> I think this is a good idea, just remove the resource specific example

These examples are retained because they are helpful for understanding, but I also added a note that they do not represent any specific resources.

AvarianKnight:
The issue here is that SQL related attacks are not related to FiveM at all.

PsychoShock:
So we just remove all the SQL section? I just added since it's something "good" for people to know. Up to you--feel free to remove what ever part you wish and merge the rest if it simplier. I'll stay available if any change.

AvarianKnight (commented):
Some suggestions based on previous conversation

PR #461, opened 2024-07-12
feat(doc/coding): new guideline

PsychoShock:
Should handle the basics on coding guidelines.

### DIFF
--- a/content/docs/developers/coding-guidelines.md
+++ b/content/docs/developers/coding-guidelines.md
@@ -39,7 +39,13 @@ Whether you're looking to improve your gameplay or contribute to the community, 
 
 ## Contribute to the project
 
-You are free to contribute on this open project, you can learn more [here](docs/contributing/how-you-can-help.md). Make sure to respect our method signature.
+We invite you to contribute to our open project! Your skills and insights are invaluable to us, and we would love for you to be part of our journey. Learn more about how you can make a difference [here](docs/contributing/how-you-can-help.md).
+
+You can also learn more about:
+- How to help with [natives](docs/scripting-manual/introduction/about-native-functions.md).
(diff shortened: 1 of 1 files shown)


colistro123 (changes requested):
Hi, I've left a few comments. Aditionally, I'm a bit confused about what you mean by 'coding guidelines' here. Generally, coding guidelines are standards to ensure code is concise and to facilitate easier collaboration among team members, given that everyone might have different coding styles.

For example, Python has the [PEP 8 coding convention](https://peps.python.org/pep-0008/). While Lua doesn't have a widely recognized style guide, LuaRocks provides its [own style guide](https://github.com/luarocks/lua-style-guide).

PsychoShock:
> Aditionally, I'm a bit confused about what you mean by 'coding guidelines' here. Generally, coding guidelines are standards to ensure code is concise and to facilitate easier collaboration among team members, given that everyone might have different coding styles.

Should I just change the structure to method signature then and move this to scripting-ref (https://github.com/citizenfx/fivem-docs/pull/463)?

colistro123:
> > Aditionally, I'm a bit confused about what you mean by 'coding guidelines' here. Generally, coding guidelines are standards to ensure code is concise and to facilitate easier collaboration among team members, given that everyone might have different coding styles.
> 
> Should I just change the structure to method signature then 

I'm not sure what you mean by 'method signature' in this context.

> move this to scripting-ref (#463)?

I think the main idea is that the title of this is wrong. It looks more like a 'Getting Started with FiveM Development' guide. I would focus on improving the current pages instead of changing the structure of the documentation, since it might be structured that way for a reason. Just a suggestion! 😊

AvarianKnight:
None of this seems to be related to coding guidelines, this should go over best practices for resources, like secure your events, check the data sent from the client, etc.

Issue #455, opened 2024-06-26
Server Setup Guides should note to disable antivirus scanning on FXServer directories


AvarianKnight:
Windows setups should probably include a note to disable Windows Defender scanning on the FXServer directory

Could also include the following example for PowerShell
`Add-MpPreference -ExclusionPath 'C:\FXServer\'`

This should fix *some* cases of players complaining about long server startups when its windows defender scanning files.

PR #467, opened 2024-08-10
Disable Antivirus scanning on FXServer directories

colistro123:
This commit adds a notice to guide users on creating an exclusion path for Microsoft Defender Antivirus to prevent it from scanning their FXServer directory.

The notice is included in the server-issues section and referenced in both the vanilla and txAdmin server setup guides.

This change addresses issue #455. However, I believe it would be more appropriately placed under the "Cfx.re Platform Server (FXServer)" section located at https://support.cfx.re/hc/en-us/sections/8856844172188-Cfx-re-Platform-Server-FXServer.


colistro123:
Rebased commits and pushed again.

PR #468, opened 2024-08-10
Update Server/Client issues section

colistro123:
This pull request removes outdated articles from the documentation that are now available in the Cfx.re support section. It also directs users to these articles under the "Help! I can't find my issue here!" header. 

Additionally, it updates certain articles in the server issues section, correcting outdated information such as the server capacity limit, which has increased from 32 to 48 slots on OneSync (up to 2048 of course), now provided by default. It also directs users to the relevant support article instructing them how to link their keymaster account to Patreon.

There's also a related PR (#436) by @PsychoShock, but it doesn't seem to address the concerns over the client-issues and server-issues sections.

This will most likely require team approval 😊


colistro123:
Re-added changes from @PsychoShock and pushed using rebase strategy.

colistro123:
For client issues, I realized that removing these sections entirely may not have been the best approach. Instead, I linked to the relevant support sections to reduce clutter on the page while still providing access to important information.

content/docs/support/server-issues.md
```diff
@@ -69,12 +69,14 @@ Using more than 48 slots requires __at least__ `Cfx.re Element Club Argentum 
 
 To use more than 48 slots follow these steps.
 
-1. Use a license key with OneSync access
+1. Use a license key with Premium access
```
AvarianKnight, line 72:
```suggestion
1. Use a license key with OneSync access
```
Probably better to not change this wording.

AvarianKnight (changes requested):

content/docs/support/server-issues.md
colistro123: > Probably better to not change this wording.

Sure! 😊

PR #473, opened 2024-08-22
fix(runtime/lua): PerformHttpRequest typo

someone005:


content/docs/scripting-reference/runtimes/lua/functions/PerformHttpRequest.md
```diff
@@ -8,7 +8,7 @@ Syntax
 ------
 
 ```lua
-PerformHttpRequest(string url, function callback[, string method = 'GET', string data = '', table headers = {}])
+PerformHttpRequest(string url, function callback[], string method = 'GET', string data = '', table headers = {})
```
AvarianKnight, line 11:
```suggestion
PerformHttpRequest(string url, function callback(number statusCode, string body, table headers, string errorData), string method = 'GET', string data = '', table headers = {}, table options = { followLocation = true })
```

AvarianKnight (changes requested):

PR #479, opened 2024-08-31
feat(loading-screens): document events

D4isDAVID:
Documents loading screen events as seen in [`LoadingScreens.cpp`](https://github.com/citizenfx/fivem/blob/56f6e3871fc376378e895a9c2d2232cc45c30e18/code/components/loading-screens-five/src/LoadingScreens.cpp), and how to use them (or at least some of them).
Not sure how correct this is since I don't have deep and extensive knowledge of the codebase.


AvarianKnight:
From a quick glance everything here looks fine, I will test this later to make sure it looks fine in the docs.

PR #481, opened 2024-09-09
docs(server/commands): Document increase_pool_size command

Nobelium-cfx:


prikolium-cfx (approved):

Identity-labs:
After testing we can't increase allowed pool size

Artifact: 10230
Repro:
increase_pool_size "InteriorProxy" 500

Error :
Requested pool size increase is invalid: requested size increase 500 for pool InteriorProxy exceeds allowed limit of 450

Why allowing to change pool size but not above actual game limit ?
Allowing us to change any pool size can help us to do testing in production environment.

Nobelium-cfx:
> Allowing us to change any pool size can help us to do testing in production environment. Why restrict to only few of them ?

The restrictions are there to make it harder for people to shoot themselves in a leg. If there is a case where higher limits are needed - we can easily adjust them.

But I understand your concern about testing. We could add a special "dangerous" flag that would remove the restrictions. But I'm not sure myself. @prikolium-cfx WDYT?

outsider31000:
every server has its own needs, I think flexibility is a must.

Identity-labs:
Redm is still at very experimental state, it's will be easy to increase some pool as testing purpose, like ped, vehicle ect ...
Where we can ask to add pool in list without creating any PR ?

prikolium-cfx:
> > Allowing us to change any pool size can help us to do testing in production environment. Why restrict to only few of them ?
> 
> The restrictions are there to make it harder for people to shoot themselves in a leg. If there is a case where higher limits are needed - we can easily adjust them.
> 
> But I understand your concern about testing. We could add a special "dangerous" flag that would remove the restrictions. But I'm not sure myself. @prikolium-cfx WDYT?

Maybe we can add startup argument `-nopoolrestriction` that will ignore our global limits on client side and some convar to do the same on server side. It will be enough for development and testing purposes. And if you need it in production you can just ask to add pool/increase pool limit in our global config, it would not take too much time for us.

Nobelium-cfx:
> Redm is still at very experimental state, it's will be easy to increase some pool as testing purpose, like ped, vehicle ect ... Where we can ask to add pool in list without creating any PR ?

Thats a good question. The pool size limits are stored in a CDN, so you won't be able to change them by creating a PR anyway.

I'm not sure. Im not sure what the best communication channel would be. If you are part of the CFX discord channel - you can write there and ping me or @prikolium-cfx . If not - you can just send a direct message on discord to one of us. My username on discord is "nobeliumcfx"

> > > Allowing us to change any pool size can help us to do testing in production environment. Why restrict to only few of them ?
> > 
> > 
> > The restrictions are there to make it harder for people to shoot themselves in a leg. If there is a case where higher limits are needed - we can easily adjust them.
> > But I understand your concern about testing. We could add a special "dangerous" flag that would remove the restrictions. But I'm not sure myself. @prikolium-cfx WDYT?
> 
> Maybe we can add startup argument `-nopoolrestriction` that will ignore our global limits on client side and some convar to do the same on server side. It will be enough for development and testing purposes. And if you need it in production you can just ask to add pool/increase pool limit in our global config, it would not take too much time for us.

Sounds good. There are higher priority tasks for me at the moment. But it should be fast to do. I'll make make sure to add the flag once I have some time.

outsider31000:
> Redm is still at very experimental state, it's will be easy to increase some pool as testing purpose, like ped, vehicle ect ... Where we can ask to add pool in list without creating any PR ?

yes its been 5 years in experimental state...
Like I said every server has its own needs allowing only a few pools servers no ones purpose...

> Maybe we can add startup argument `-nopoolrestriction` that will ignore our global limits on client side and some convar to do the same on server side. It will be enough for development and testing purposes. And if you need it in production you can just ask to add pool/increase pool limit in our global config, it would not take too much time for us.

what would be the point in testing things if we can't use in production? most tests are done in production specially testing and seeing pool sizes, and then we  need to *ask* and wait, and wait to be in release version to use it. that's a lot of steps.
 
The flag is a good idea but I think its unacessery code that will stay forever in the code base. *not that is any of my concern just an observation*

**in my opinion** just don't restric it, it should be the server responsability on how they use it.
can't make everything fool proof if you know what I mean .

Korioz:
You'll not have to wait for it to be in release version, the list is the same for all versions.

This is recent so yes the pool restrictions list will be edited a few times but over time the list will be more and more stable, if someone need a limit to be adjusted he will request Cfx team once he proves this is needed.

Something that is missing is the information of how to request adjustments of the limits, this should be notified in the console when hitting a pool restriction so anyone that needs to up the limit will know how he should request it.

iafondo:
First of all thankyou i was stuck on 2612 for a while, this allows me to go current and give me 15% room to play so any improvements are appreciated, and like said before if possible to give responsibility to the server to test some pools would be great

outsider31000:
thank you 
https://github.com/citizenfx/fivem/issues/2888#issuecomment-2483942993 
@Nobelium-cfx  could you please have a look at this.
this was requested some time ago

kamelzarandah:
is increasing the pool limit for networked objects possible ? right now once you pass that limit the object automatically despawns

outsider31000:
I believe you can already do that see: https://github.com/citizenfx/fivem-docs/pull/507/files although is not advised to use in production you can.

SevenLife:
No, we already tried for these issue sadly :(

Nobelium-cfx:
> is increasing the pool limit for networked objects possible ? right now once you pass that limit the object automatically despawns

@kamelzarandah  Maybe. Can you give me a bit more context?

> On canary with latest artifacts? not sure if it's in production

@outsider31000 It is only available on Canary right now. It will get to beta and production with the regular beta/prod push.

> No, we already tried for these issue sadly :(

@SevenLife Did you increase the pool but it didn't help? Or you couldn't increase the pool size at all? 

If its the first - then the pool size increase can't help the issue at all and we should search for some other ways to deal with it.

If the second - please let me know, I should take a look into it then.

> I believe you can already do that see: https://github.com/citizenfx/fivem-docs/pull/507/files although is not advised to use in production you can.

Yes, please don't use it in production.

Instead, if you test the size increase with it and it works for you and resolves the problem - then its much easier for me to just add it to allow list because I know that it indeed worksw

SevenLife:
Thanks Nobelium, we will run some test soon, it's more likelly @zetafe1 how have found this issue but this will impact all server in the futur for sure.

PR #483, opened 2024-09-27
Improved specifications for loadscreen

Maximus7474:
Currently the deferrals.handover method is only shown to work in a lua example however it isn't specified that it's only a feature in the lua runtime and not for example the JS runtime.


Maximus7474:
Please ignore, there was a typo in the project.

PR #484, opened 2024-09-29
chore(build): add a script to build the docs

ahcenezdh:
Add a script to build and run hugo docs, the script is checking if the user have GCC and Go installed on his computer, and if the user don't have hugo then it will build it and run it, if the user already have hugo it will directly run it (the script don't make the difference between the official hugo and the one from dbub)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 /node_modules
 
 .vscode
+hugo.exe
+/hugo
--- a/build.bat
+++ b/build.bat
@@ -1,0 +1,70 @@
+@echo off
+setlocal enabledelayedexpansion
+
+:check_go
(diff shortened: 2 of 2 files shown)


AvarianKnight:
I think it would be best to wait on this for the moment, hopefully we won't need a custom build in the future.

ahcenezdh:
> I think it would be best to wait on this for the moment, hopefully we won't need a custom build in the future.

No problem, i'm going to let this open in case someone want to build the documentation for now

AvarianKnight:
We should probably (even if temporarily) add this in to make setting up hugo easier, the RenderShortcode changes I PR'd had a few changes that break some styling so its not viable to merge it.

PR #486, opened 2024-10-04
tweak: swap to RenderShortcodes and use latest version of Hugo

AvarianKnight:
This supersedes https://github.com/citizenfx/fivem-docs/pull/484


technetium-cfx:
Fixed in 5d44bf5295f043bc8d455c3797c95baaa77bf60e

PR #494, opened 2024-10-28
document speech params and names

coalaura:
see https://github.com/citizenfx/natives/pull/1217#issuecomment-2442782068


content/docs/game-references/speeches.md
```diff
@@ -0,0 +1,554 @@
...
+    "SPEECH_PARAMS_SHOUTED",
+    "SPEECH_PARAMS_STANDARD"
+}
+```
+
+Speech Names
+--------
+
+```cpp
+static const char *SpeechNames[] = {
```
colistro123, line 65: Isn't it better to provide arrays that match the current programming languages (C#, Lua and JS) that way people can re-use them if needed? Just my two cents 😊

coalaura: Since the native documentation uses cpp for all enum or list codeblocks, i figured it'd make sense to continue the trend here

coalaura: And since string arrays are (to my knowledge) defined basically the same in lua, c# and js (with the last 2 using [ instead of {), so you should be able to easily reuse them, if there is a need to copy the list entirely

colistro123 (approved):
LGTM at a glance.

content/docs/game-references/speeches.md
```diff
@@ -0,0 +1,554 @@
+---
+title: Speeches
+weight: 770
```
PsychoShock, line 3: Please remove this, this is auto-generated. Not really even used by Hugo (besides priority iirc).

content/docs/game-references/speeches.md
```diff
@@ -0,0 +1,554 @@
...
+
+Example
+-------
+
+### Lua
+
+``` lua
+-- Makes the ped shout an insult
+PlayPedAmbientSpeechNative(ped, "GENERIC_INSULT_HIGH"--[[speechName]], "SPEECH_PARAMS_FORCE"--[[speechParam]])
+```
```
PsychoShock, line 14: Never forget people will copy / paste and expect the "example" to work. You shoud define `ped`.

content/docs/game-references/speeches.md
```diff
@@ -0,0 +1,554 @@
...
+
+Speech Params
+------------
+
+```cpp
+static const char *SpeechParams[] = {
```
PsychoShock, line 20: The result won't be "good", nobody needs to declare all of them in a code. They should be set in a table like https://github.com/citizenfx/fivem-docs/blob/master/content/docs/game-references/game-events.md and people will have the possibilibity to set a description (it makes easier to people to know what they are in the future after)

content/docs/game-references/speeches.md
colistro123: Not sure what is meant by this, `weight` is used to figure out where it should be placed in the navigation bar (the priority you're mentioning).

The title is also applied, see [Vehicle Models](https://docs.fivem.net/docs/game-references/vehicle-models/).

https://github.com/citizenfx/fivem-docs/blob/master/content/docs/game-references/vehicle-models.md?plain=1#L2-L3

PsychoShock: Yeah but it shouldnt be a "random" number set by the PR. I guess here priority doesnt "mean" anything, so any number is "okay" I suppose.

colistro123: Ah I understand what you mean now. Anyway, if this is meant to be at the bottom of 'Game references', it should be `790` or anything after `780` which is basically 'Vehicle Colours'.

coalaura: Apologies, i thought it was set manually so i tried putting it where i felt made sense. Will remove

content/docs/game-references/speeches.md
coalaura: The speech natives (from my experience) don't seem to really work on player peds, so there would need to be a decent bit of "boilerplate" code to find a working ped. Unless you have a different idea?

content/docs/game-references/speeches.md
coalaura: Sounds good! Will update it :)

content/docs/game-references/speeches.md
colistro123: > Apologies, i thought it was set manually so i tried putting it where i felt made sense. Will remove

This wasn't meant to be removed. I believe PsychoShock was merely pointing out that [weight](https://gohugo.io/methods/page/weight/) will be used for priority. Said declaration such as 'title' and others should be present in all markdown files. I recommend you to check how the current documentation is written in case of any doubts.

coalaura: > This wasn't meant to be removed. I believe PsychoShock was merely pointing out that [weight](https://gohugo.io/methods/page/weight/) will be used for priority. Said declaration such as 'title' and others should be present in all markdown files. I recommend you to check how the current documentation is written in case of any doubts.

Was going off of this yes

coalaura: i'll add the title back

content/docs/game-references/speeches.md
PsychoShock: Does it not work directly on a random created ped?

AvarianKnight: Just use https://docs.fivem.net/natives/?_0xB4AC7D0CF06BFE8F to create a random ped at the players current position

coalaura:
Ok i believe i've found where the speeches are stored. They seem to be stored in `x64\audio\audio_rel.rpf\config\game.dat151.rel`, so i'll update my list with all the ones listed in there

content/docs/game-references/speeches.md
coalaura: Updated the example

PR #495, opened 2024-10-31
tweak(onesync): update docs to make use of the orphan natives instead of culling natives

AvarianKnight:


content/docs/scripting-reference/onesync/_index.md
```diff
@@ -140,30 +146,55 @@ The original examples can be found in the following forum [post][original-scope-
...
+-- creates the ped at the same coords!
+local ped = CreatePed(4, GetHashKey("a_m_y_acult_01"), 2204.795, -887.9213, 1461.224, 90.0, true, true)
+-- guarantee that the server created entity will be persistent for the server
+SetEntityOrphanMode(ped, 2)
 ```
 
 ## I want persistent entities, how do I do it?
-Create them server-side as shown up above. If you want to stop this entity from 'despawning' when far away, you could always set the entity culling radius to a very large number by using the following native: `SetEntityDistanceCullingRadius`. You should use this sparingly though, and you should rather do manual lower-level sync such as latent events for blips if you have large amounts of entities. Make sure to make use of network id's since they remain valid as long as the entity is alive.
+If you want to guarantee an entity will not be removed by the server you should use {{% native_link "SET_ENTITY_ORPHAN_MODE" %}} with the 'KeepEntity' flag.
+This will guarantee that the server will not delete the vehicle, though the client will still be able to.
```
AvarianKnight, line 165:
```suggestion
This will guarantee that the server will not delete the vehicle, but the client will still be able to request its deletion.
```

PR #499, opened 2024-11-07
Correct broken URLs

colistro123:
Correct broken URLs without preceding '/' which ultimately lead to a 404 page.


AvarianKnight:
LGTM, Thanks.

PR #502, opened 2024-11-15
docs(resource-manifest): include oal information

Z3rio:
Hello, I've noticed that there's genuinely no clear information on OAL in the documentation

I do understand that this might be an experimental feature, but a lot of people want to understand it and start to play around with it.
A good start would be to add it to the resource manifest docs, but with a disclaimer.

### DIFF
--- a/content/docs/scripting-reference/resource-manifest/resource-manifest.md
+++ b/content/docs/scripting-reference/resource-manifest/resource-manifest.md
@@ -308,7 +308,7 @@ use_experimental_fxv2_oal "yes"
 ```
 
 {{% alert color="warning" %}}
-This feature is still experimental and also **requires** [Lua 5.4](#lua54) to be used.
+This feature is still experimental and **requires** [Lua 5.4](#lua54) to be used.
 {{% /alert %}}
 
 #### Vector unpacking
--- a/content/docs/scripting-reference/resource-manifest/resource-manifest.md
+++ b/content/docs/scripting-reference/resource-manifest/resource-manifest.md
@@ -299,6 +299,18 @@ Marks the current resource as a replacement for the specified resource. This mea
 provide 'mysql-async'
 ```
 
+### use_experimental_fxv2_oal
+
+This will enable the usage of OAL for Lua. This aims to correct return-types of natives and provide better performance via faster native calls.
+
+```lua
+use_experimental_fxv2_oal "yes"
+```
+
+{{% alert color="warning" %}}
+This feature is still experimental, it also **requires** Lua 5.4 to be used.
+{{% /alert %}}
+
 ### clr_disable_task_scheduler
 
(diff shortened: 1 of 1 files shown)


Z3rio:
@AvarianKnight Does this look better? 🚀

content/docs/scripting-reference/resource-manifest/resource-manifest.md
```diff
@@ -299,6 +299,37 @@ Marks the current resource as a replacement for the specified resource. This mea
...
+### use_experimental_fxv2_oal
+
+This will enable the usage of OAL (One Argument List) for Lua. This aims to correct return-types of natives and provide better performance via faster native calls.
+
+```lua
+use_experimental_fxv2_oal "yes"
+```
+
+{{% alert color="warning" %}}
+This feature is still experimental and also **requires** [Lua 5.4](#lua54) to be used.
```
AvarianKnight, line 311:
```suggestion
This feature is still experimental and **requires** [Lua 5.4](#lua54) to be used.
```

content/docs/scripting-reference/resource-manifest/resource-manifest.md
```diff
@@ -299,6 +299,37 @@ Marks the current resource as a replacement for the specified resource. This mea
...
+{{% /alert %}}
+
+#### Vector unpacking
+
+Vector unpacking is not a feature when using OAL, this means that you will have to explicitly unpack each coordinate into the native. Example below.
```
AvarianKnight, line 316:
```suggestion
Vector unpacking does not work when using OAL, this means that you will have to manually unpack coordinates instead of providing the `vector3`.
```

content/docs/scripting-reference/resource-manifest/resource-manifest.md
```diff
@@ -299,6 +299,37 @@ Marks the current resource as a replacement for the specified resource. This mea
...
+SetEntityCoords(ped, coords.x, coords.y, coords.z) -- works both with OAL, and without
+```
+
+#### Downsides
+
+Using OAL does have its downsides, if used incorrectly. Which is especially easy to do in more unexplored/unknown areas, such as RedM.
+
+More specifically, if a native is called with incorrect parameters. Such as unpacked vectors, `0` or `1` instead of `false` or `true`, and so on.
+
+Therefore, please ensure the types of the native parameters if you want to use OAL.
```
AvarianKnight, line 331:
```suggestion
OALs main downside is that it cannot be used if the parameter type is wrong, as internally it will be converted to whatever the underlying type is.

This means that for natives that are undocumented or don't have the right types OAL will break in unexpected ways, or likely just not working at all.
```

AvarianKnight (changes requested):

AvarianKnight:
LGTM, Thanks 😃

PR #504, opened 2024-11-19
fix: Change Heading Licenses to Identifiers

TheGamerzs:
Change to the correct name
Used as a referenced URL as identifier here: https://github.com/citizenfx/fivem/pull/2948


AvarianKnight:
LGTM, Thanks.

PR #508, opened 2024-11-27
tweak(scripting/listening-for-events): Source - mention asynchronous

PulseHarbor:
Add usage about `source`, since its affected during an asynchronous thread.
Happens in Lua; This probably happens with JS as-well, based on JS thread knowledge.

### DIFF
--- a/content/docs/scripting-manual/working-with-events/listening-for-events.md
+++ b/content/docs/scripting-manual/working-with-events/listening-for-events.md
@@ -5,7 +5,7 @@ weight: 431
 
 To use existing events in your resources, you need to listen for them. This is the same for client/server side scripts.
 
-In Lua/JS, the `source` variable (global) will contain the player ID that triggered the event. You should most likely save it in a local variable if you're going to use it after the event returns or the event contains asynchronous functionality.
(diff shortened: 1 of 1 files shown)


content/docs/scripting-manual/working-with-events/listening-for-events.md
```diff
@@ -5,7 +5,7 @@ weight: 431
 
 To use existing events in your resources, you need to listen for them. This is the same for client/server side scripts.
 
-In Lua/JS, the `source` variable (global) will contain the player ID that triggered the event. You should most likely save it in a local variable if you're going to use it after the event returns.
+In Lua/JS, the `source` variable (global) will contain the player ID that triggered the event. You should most likely save it in a local variable if you're going to use it after the event returns or the event contains asynchronous functionality.
```
colistro123, line 8: What is meant by asynchronous functionality here? Does it mean that the event is classified as a `Task` in `C#`, that it uses a `coroutine` in `lua` i.e. `Citizen.CreateThread`? I believe it needs to be a little bit more explicit and explain as on to what exactly is meant by this, as in what scenarios would actually cause this to happen when directly dealing with `source`.

AvarianKnight, line 8:
```suggestion
In Lua/JS, the `source` variable (global) will contain the player ID that triggered the event. You must save the source to a local variable if you plan on using it after the event returns, or inside of an asynchronous scope, as the source is only guaranteed to be valid for the initial event call. Calling `Wait` in Lua or awaiting something in JS will cause the global source to be set back to default.
```

This should probably provide enough context on why you should do this

AvarianKnight (changes requested):

AvarianKnight:
LGTM, Thank you :)

PR #513, opened 2024-12-17
feat: Add Local TLS Proxy documentation

FabianTerhorst:
This adds documentation in case of an local TLS proxy is required to improve handling of TLS attacks in case services like Cloudflare are not usable.


AvarianKnight (approved):
LGTM

PR #515, opened 2025-01-02
tweak(console-commands): remove 'net_maxPackets' as it no longer exists

william-des:
This PR removes the `net_maxPackets` command from the 'console commands' documentation.

When attempting to execute this command, an error message is displayed indicating that the command does not exist. After investigating the codebase, I could not find any trace of this command in the current version of the code.


AvarianKnight:
Looks like this hasn't existed for a while.

Thanks for the fix

PR #529, opened 2025-02-28
docs(server/commands): update CMoveObject value & add new option for fragGtaInst

Legacy-TacticalGamingInteractive:
Update value for `CMoveObject` as per posted limit. Reference: https://github.com/citizenfx/fivem/issues/2492#issuecomment-2569088478

Update: Also added `fragInstGta`


Legacy-TacticalGamingInteractive (commented):
fragGtaInst added by @prikolium-cfx for reference: https://github.com/citizenfx/fivem/issues/3290

PR #546, opened 2025-07-03
Lua 5.3 deprecation & NodeJS 22

nimoalr:
Update the scripting manual and resource manifest to mention Lua 5.3 deprecation and the optionally available NodeJS version 22 on fxserver.

### DIFF
--- a/content/docs/scripting-reference/resource-manifest/resource-manifest.md
+++ b/content/docs/scripting-reference/resource-manifest/resource-manifest.md
@@ -286,7 +286,7 @@ The valid constraint types are as follows:
 
 Enables Lua 5.4. You can read more about Lua 5.4 at http://www.lua.org/manual/5.4/manual.html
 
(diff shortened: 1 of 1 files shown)


ook3D:
Is it intentional that it says June 2023 was when Lua 5.3 was deprecated?

AvarianKnight:
Is there any reason to even keep mentions to Lua 5.3 at all, even the deprecation date? It all seems like useless information instead of just completely removing it.

ook3D:
> Is there any reason to even keep mentions to Lua 5.3 at all, even the deprecation date? It all seems like useless information instead of just completely removing it.

i agree, it should simply mention only 5.4 is supported

nimoalr:
> Is there any reason to even keep mentions to Lua 5.3 at all, even the deprecation date? It all seems like useless information instead of just completely removing it.

My goal was to redirect users to the content of the announcement stating what are the differences between the 5.3 and 5.4 runtimes without needing to paste all of the differences directly in the docs. Do you recommend changing the wording?

PR #548, opened 2025-07-11
Explain how to enable debug information

colistro123:
This commit explains how to enable embedded debug information on .NET assemblies to ensure the Mono scripting runtime displays stack traces correctly. 

The approach is based on the discussion found in the forum post: https://forum.cfx.re/t/c-deferral-server-crash/119960/2


avail:
pdb2mdb also works (and adding the generated .mdb to files array)

colistro123:
> pdb2mdb also works (and adding the generated .mdb to files array)

Nice to know, thanks! I'll test that tomorrow so I can add it, kind of tired by now 😊

PR #581, opened 2026-04-24
Add new entry for mp2026_patch_01 in server commands

davismasondaniel:
| 3788   | mp2026_patch_01                            | A Safehouse in the Hills Patch_2026_1   |


technetium-cfx:
Patches are not included in this list as it only pertains content level updates

PR #583, opened 2026-05-27
Improve the Migrating from Other Platforms page

Gogsi:
- Detailed description of Server RPCs and Server setters with pros and cons
- Add ox_lib Zones as an alternative to Polyzone
- Add links to other relevant docs pages

### DIFF
--- a/content/docs/scripting-manual/migrating-from-other-platforms/_index.md
+++ b/content/docs/scripting-manual/migrating-from-other-platforms/_index.md
@@ -5,11 +5,79 @@ weight: 95
 
 This page covers the most common topics and differences when migrating from other GTAV multiplayer modification platforms to FiveM.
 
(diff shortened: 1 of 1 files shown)


Gogsi:
Open to feedback on the RPC/setter talk. 

The code I have written for server setters is technically not enough to be 100% safe. As mentioned in the warning, if the owner migrates before the CVehicleAppearanceDataNode syncs, the changes will be lost. You would have to do something like qbx_core and [monitor for if the changes successfully applied server-side ](https://github.com/Qbox-project/qbx_core/blob/12d2cae243b29efaf4ea9cba22e7362fca65eb9f/modules/lib.lua#L351), but I didn't want to make the example too large so I just added a warning. This code should work *most* of the time at least.

PR #586, opened 2026-05-31
Clarify Node.js 22 server thread affinity

Felipellira:
Clarifies that the JavaScript thread-affinity workaround documented for Node.js callbacks applies to the default Node.js 16 server runtime.

Adds a note that resources using `node_version '22'` no longer need to wrap server natives or event calls in `setImmediate` solely after Node.js I/O callbacks or awaited Node.js operations.

### DIFF
--- a/content/docs/scripting-manual/runtimes/javascript.md
+++ b/content/docs/scripting-manual/runtimes/javascript.md
@@ -108,7 +108,9 @@ Gotchas
 -------
 
 ### Thread affinity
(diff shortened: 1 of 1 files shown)


Felipellira:
I came to this conclusion after thoroughly testing the issue in various scenarios; when I couldn’t reproduce it, I went straight to the source code, analyzed it, and ran Codex + Claude to see if I’d missed anything, and it turns out my hunch was correct.

AvarianKnight:
Was this still even a thing in node 16? This might just be outdated in general.

PR #589, opened 2026-06-16
fix(nui): replace workaround with nui callback native

Hellslicer:


content/docs/scripting-manual/nui-development/nui-callbacks.md
```diff
@@ -3,20 +3,14 @@ title: NUI callbacks
 weight: 20
 ---
 
-NUI can also send calls back to the game using so-called 'NUI callbacks'. These are currently only fully supported in
-Lua, other languages can be used but need a bit of a [tricky workaround][workaround] as these predate function
-references in codegen.
+NUI can also send calls back to the game using so-called 'NUI callbacks'. These are currently fully supported in
+Lua and other languages.
```
AvarianKnight:
```suggestion
NUI can also send calls back to the game using NUI callbacks.  
```

The extra tidbits here are un-needed

content/docs/scripting-manual/nui-development/nui-callbacks.md
```diff
@@ -3,20 +3,14 @@ title: NUI callbacks
...
 
-<!-- #GAMETODO: actually fix that? -->
-
-Generally, you'll use the [RegisterNUICallback][registernuicallback] function in Lua, and the
-{{% native_link "REGISTER_NUI_CALLBACK_TYPE" %}} native along with an event handler in other languages.
-
-Both work very similarly, and we'll describe both below:
+You can register NUI callbacks using {{% native_link "REGISTER_NUI_CALLBACK" %}} native, and if using Lua, the old convenience wrapper [RegisterNUICallback][registernuicallback].
```
AvarianKnight:
```suggestion
You can register NUI callbacks using the {{% native_link "REGISTER_NUI_CALLBACK" %}} native.
```

We shouldn't recommend using the old way

content/docs/scripting-reference/runtimes/lua/functions/RegisterNUICallback.md
```diff
@@ -2,9 +2,52 @@
...
+        cb({ error = 'No such item!' })
+        return
+    end
+
+    -- and so does callback response data
+    cb(itemCache[itemId])
+end)
 ```
+
+Browser-side JS (NUI PART)
```
AvarianKnight:
```suggestion
Browser-side JS, this calls into the NUI Callback to and will get the NUI data back as a response
```

AvarianKnight (changes requested):

content/docs/scripting-reference/runtimes/lua/functions/RegisterNUICallback.md
```diff
@@ -2,9 +2,52 @@
 title: RegisterNUICallback
 ---
 
+[Old function][oldfunction] to register NUI (NewUI) callbacks, kept for compatibility reasons.
```
technetium-cfx: Would recommend using an alert or similar to call out that this is a legacy function kept for compatibility, and that the recommended way is to use the native

Something like this could work
```
{{% alert color="warning" %}}
**Legacy API:** This function is kept for backwards compatibility. New code should use the {{% native_link "REGISTER_NUI_CALLBACK" %}} native instead. Refer to the current [NUI callback documentation][nuicallbackinlua] for examples.
{{% /alert %}}
```

technetium-cfx (changes requested):

PR #591, opened 2026-06-25
docs (scripting-reference) - routing bucket events

xalva98:
Adds some undocumented events about routing buckets


AvarianKnight:
LGTM

PR #595, opened 2026-07-21
Add light/dark mode support

TGIANN:
I added and tested the dark mode feature with Claude's help. Both dark mode and light mode work correctly on every page. However, the old FiveM theme was set as the default theme; I'm not sure why that happened

image

image

image

image


PsychoShock:
Thank Claude, it will finally not hurt my eyes anymore 🙏

TGIANN:
I guess even Claude didn't stop us from being blind

PR #599, opened 2026-07-27
Document NUI devtools port for FiveM for GTAV Enhanced

TheGamerzs:
The Fullscreen NUI page states that CEF remote debugging tools are always exposed on `http://localhost:13172/`. That is correct for FiveM Legacy and RedM, but FiveM for GTAV Enhanced exposes them on `http://localhost:3232/` instead.

This replaces the single hardcoded address with a small table covering both clients.


technetium-cfx:
This will be changed to match in a future patch. Related: https://github.com/citizenfx/rfc/discussions/184

PR #600, opened 2026-07-30
Add getPlayers JS documentation

sw4m:
Added documentation to the obscure JS equivalent of Lua's GetPlayers.


AvarianKnight:
A warning should probably be put here that this returns a string representation of the source, not an integer representation.

sw4m:
Good idea, I have specified that they are returned as strings

sw4m:
@AvarianKnight good to go?

AvarianKnight:
LGTM. Thanks for your contribution 😄

